PR #4, opened 2017-03-06
[MLP-117] Integrate Autofac

mvazquez-ms:
## Background

MvvmCross default container is very limited. Autofac provides much more flexibility and has multi-platform support.

## Changes done

- Replaced MvvmCross default container with Autofac
- Added .NET Standard reference on Android-specific assembly

### DIFF
--- a/Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
+++ b/Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
@@ -1,7 +1,11 @@
 using Android.Content;
+using Autofac;
+using Autofac.Extras.MvvmCross;
 using MvvmCross.Core.ViewModels;
 using MvvmCross.Droid.Platform;
+using MvvmCross.Platform.IoC;
 using MvvmSeed.Application;
+using MvvmSeed.Application.Modules;
 
 namespace MvvmSeed.Android
 {
@@ -9,9 +13,21 @@ namespace MvvmSeed.Android
     {
         public Setup(Context applicationContext) : base(applicationContext) { }
 
-        protected override IMvxApplication CreateApp()
+        protected override IMvxApplication CreateApp() => new App();
+
+        protected override IMvxIoCProvider CreateIocProvider()
         {
-            return new App();
+            var containerBuilder = new ContainerBuilder();
+            containerBuilder.RegisterModule<ApplicationModule>();
+
+            // This is an important step that ensures all the ViewModel's are loaded into the container. Without it, MvvmCross wouldn't register them by itself; needs more investigation.
+            var viewModelsAssembly = typeof(App).Assembly;
+            containerBuilder.RegisterAssemblyTypes(viewModelsAssembly)
+                .AssignableTo<MvxViewModel>()
+                .As<IMvxViewModel, MvxViewModel>()
+                .AsSelf();
+
+            return new AutofacMvxIocProvider(containerBuilder.Build());
         }
     }
 }
--- a/Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/packages.config
+++ b/Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/packages.config
@@ -1,7 +1,55 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="Autofac" version="4.4.0" targetFramework="monoandroid71" />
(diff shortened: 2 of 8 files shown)


Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
```diff
@@ -1,17 +1,33 @@
...
-        protected override IMvxApplication CreateApp()
+        protected override IMvxApplication CreateApp() => new App();
+
+        protected override IMvxIoCProvider CreateIocProvider()
         {
-            return new App();
+            var containerBuilder = new ContainerBuilder();
+            containerBuilder.RegisterModule<ApplicationModule>();
+
+            // This is an important step that ensures all the ViewModel's are loaded into the container. Without it, MvvmCross wouldn't register them by itself; needs more investigation.
```
AlphaGit, line 23: The "needs more investigation" part confuses me a little bit. Is this not acceptable?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/packages.config
```diff
@@ -1,7 +1,55 @@
...
+  <package id="System.Security.Cryptography.Primitives" version="4.3.0" targetFramework="monoandroid71" />
+  <package id="System.Security.Cryptography.X509Certificates" version="4.3.0" targetFramework="monoandroid71" />
+  <package id="System.Text.Encoding" version="4.3.0" targetFramework="monoandroid71" />
+  <package id="System.Text.Encoding.Extensions" version="4.3.0" targetFramework="monoandroid71" />
+  <package id="System.Text.RegularExpressions" version="4.3.0" targetFramework="monoandroid71" />
+  <package id="System.Threading" version="4.3.0" targetFramework="monoandroid71" />
+  <package id="System.Threading.Tasks" version="4.3.0" targetFramework="monoandroid71" />
+  <package id="System.Threading.Timer" version="4.3.0" targetFramework="monoandroid71" />
+  <package id="System.Xml.ReaderWriter" version="4.3.0" targetFramework="monoandroid71" />
+  <package id="System.Xml.XDocument" version="4.3.0" targetFramework="monoandroid71" />
```
AlphaGit, line 54: Are all of these required by Autofac?

AlphaGit (approved):

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
mvazquez-ms: Well, supposedly, MvvmCross is already doing that under the hood.. but it's not.  I would like to make sure there are no 'double registrations' somewhere.. I remember reading about it on a discussion somewhere.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/packages.config
mvazquez-ms: Not directly.. but yes. I'm using an Autofac version that is targeting .NET Standard, and this is what you get when referencing .NET standard package (it's distributed via Nuget).

mvazquez-ms: Check [this](https://www.nuget.org/packages/NETStandard.Library/).

AlphaGit: Thanks!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
AlphaGit: Got it -- thanks!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Modules/ApplicationModule.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Autofac;
+using MvvmCross.Core.ViewModels;
+using MvvmSeed.Application.ViewModels;
+
+namespace MvvmSeed.Application.Modules
+{
+    public class ApplicationModule : Module
```
matiasbeckerle, line 7: In the near future, if you are going to add another Module consider adding a description comment on each header (of what should be registered in that module). I've found it very useful in CS3 for example. Just an idea.

matiasbeckerle (approved):

PR #5, opened 2017-03-06
[MLP-119] Integrate RealmDB

mvazquez-ms:
## Background

One of the objectives of this seed project is to provide a simple local storage mechanism. The obvious choices for mobile applications are Sqlite and Realm.

This PR shows how to integrate Realm and store/retrieve objects from it. The code is simple and maybe a bit stupid, but it's enough to show how Realm works.

## Changes Done

- Referenced Realm packages 
- Configured Realm weaver (yep, it relies on IL weaving to add all the boilerplate code within the build process, I've check the Fody module used for this and it adds extra instructions to classes inheriting from `RealmObject`)
- Created a Realm Entity called `RandomizedStringEntity `
- Created `IStringRandomizerService` and its implementation relying on Realm db

## Notes

I like it, I think it might be a better choice than Sqlite, but I would like to play a bit more with it. This is for early feedback.

One drawback is that  `Realm` class doesn't implement an interface, so a framework other than Moq will be required for UnitTests (one capable of mockings non virtual methods, like TypeMock).

### DIFF
--- a/Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/FodyWeavers.xml
+++ b/Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/FodyWeavers.xml
@@ -1,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Weavers>
+  <RealmWeaver/>
+</Weavers>
--- a/Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Services/StringRandomizerService.cs
+++ b/Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Services/StringRandomizerService.cs
@@ -1,0 +1,53 @@
+﻿using MvvmSeed.Application.Data;
+using MvvmSeed.Domain.Services;
+using Realms;
+using System;
+using System.Linq;
+
+namespace MvvmSeed.Application.Services
+{
+    /// <summary>
+    /// Simple implementation of <see cref="IStringRandomizerService"/> that stores randomization info on a local database (Realm)
+    /// </summary>
+    public class StringRandomizerService : IStringRandomizerService
+    {
+        //TODO - When attempting to write UTs for this service, found out I couldn't because Realm class doesn't implement an interface/virtual methods. 
+        // Evalute alternatives (weaving Realm class, Moq alternatives, abstract Realm into ILocalStorage, etc)
+        private readonly Realm _realmDb;
+
+        public StringRandomizerService(Realm realmDb)
+        {
+            _realmDb = realmDb;
+        }
+
+        public string RandomizeString(string input)
+        {
+            var random = new Random();
+            var randomizedString = new string(input.ToCharArray().OrderBy(s => random.Next(2) % 2 == 0).ToArray());
+
+            var lastRandomizedString = _realmDb.Find<RandomizedStringEntity>(RandomizedStringEntity.DefaultRandomizedStringPrimaryKey) ?? new RandomizedStringEntity { Id = RandomizedStringEntity.DefaultRandomizedStringPrimaryKey };
+
+            //TODO - Concurrency tests when writing to DB
+            using (var transaction = _realmDb.BeginWrite())
+            {
+                lastRandomizedString.LastTransformationTimestamp = DateTimeOffset.UtcNow;
+                lastRandomizedString.SampleTransformationCount++;
+                lastRandomizedString.LastTransformationValue = randomizedString;
+                _realmDb.Add(lastRandomizedString, update: true);
+                transaction.Commit();
+            }
+            return randomizedString;
+        }
+
+        public DateTimeOffset LastRandomizationTimestamp => GetLastEntryFromDatabse()?.LastTransformationTimestamp ?? DateTimeOffset.MinValue;
+
+        public int RandomizationsCount => GetLastEntryFromDatabse()?.SampleTransformationCount ?? 0;
+
+        public string LastRandomizedValue => GetLastEntryFromDatabse()?.LastTransformationValue ?? "Hello World!";
+
+        private RandomizedStringEntity GetLastEntryFromDatabse()
+        {
+            return _realmDb.Find<RandomizedStringEntity>(RandomizedStringEntity.DefaultRandomizedStringPrimaryKey);
+        }
+    }
+}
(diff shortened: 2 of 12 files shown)


Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/FodyWeavers.xml
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Weavers>
+  <RealmWeaver/>
```
briansztamfater, line 3: What is the purpose of this file and line?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Services/StringRandomizerService.cs
```diff
@@ -0,0 +1,53 @@
...
+            return randomizedString;
+        }
+
+        public DateTimeOffset LastRandomizationTimestamp => GetLastEntryFromDatabse()?.LastTransformationTimestamp ?? DateTimeOffset.MinValue;
+
+        public int RandomizationsCount => GetLastEntryFromDatabse()?.SampleTransformationCount ?? 0;
+
+        public string LastRandomizedValue => GetLastEntryFromDatabse()?.LastTransformationValue ?? "Hello World!";
+
+        private RandomizedStringEntity GetLastEntryFromDatabse()
```
matiasbeckerle, line 48: Typo: GetLastEntryFromData**bse**

matiasbeckerle (approved):
Looking fine but just want to remark two things:

- Testing should be addressed somehow.
- Not sure how much useful but maybe you can place db related code in other layer instead of doing the transactions directly on the services. What do you think?

briansztamfater (approved):

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/FodyWeavers.xml
mvazquez-ms: (Damn, I knew someone would ask about it..  😄 )

The complete answer is long.. but this is the configuration file for Fody, which is a tool to perform IL weaving (add extra code as part of the build process).

The common usage for IL weaving is to avoid boilerplate code..  so, for example, all the `INotifyPropertyChanged` boilerplate code to notify property changes and event definitions could be added automatically as part of the build process, as long as you can identity which properties need notification. That will save considerable time for developers and will keep the code clean.

The weaver for Realm looks something like this:

```
private void ReplaceGetter(PropertyDefinition prop, string columnName, MethodReference getValueReference)
    {
        // A synthesized property getter looks like this:
        //   0: ldarg.0
        //   1: ldfld <backingField>
        //   2: ret
        // We want to change it so it looks like this:
        //   0: ldarg.0
        //   1: call Realms.RealmObject.get_IsManaged
        //   2: brfalse.s 7
        //   3: ldarg.0
        //   4: ldstr <columnName>
        //   5: call Realms.RealmObject.GetValue<T>
        //   6: ret
        //   7: ldarg.0
        //   8: ldfld <backingField>
        //   9: ret
        // This is roughly equivalent to:
        //   if (!base.IsManaged) return this.<backingField>;
        //   return base.GetValue<T>(<columnName>);

        var start = prop.GetMethod.Body.Instructions.First();
        var il = prop.GetMethod.Body.GetILProcessor();

        il.InsertBefore(start, il.Create(OpCodes.Ldarg_0)); // this for call
        il.InsertBefore(start, il.Create(OpCodes.Call, _references.RealmObject_get_IsManaged));
        il.InsertBefore(start, il.Create(OpCodes.Brfalse_S, start));
        il.InsertBefore(start, il.Create(OpCodes.Ldarg_0)); // this for call
        il.InsertBefore(start, il.Create(OpCodes.Ldstr, columnName)); // [stack = this | name ]
        il.InsertBefore(start, il.Create(OpCodes.Call, getValueReference));
        il.InsertBefore(start, il.Create(OpCodes.Ret));

        Debug.Write("[get] ");
    }
```

At the same time, it performs some 'static analysis'. for example, if you declare a `DateTime` property that should be stored on the database, it will fail, since Realm used `DateTimeOffset`, which is more appropriate to handle dates on a db.

So, Fody is a tool for IL weaving, and on this file I'm configuring Realm weaver to do its 'magic' on the build process. This file can contain multiple weavers.. 

If this is 'clean' or not, it's another discussion.. but it is how Realm works on .NET.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Services/StringRandomizerService.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Linq;
+
+namespace MvvmSeed.Application.Services
+{
+    /// <summary>
+    /// Simple implementation of <see cref="IStringRandomizerService"/> that stores randomization info on a local database (Realm)
+    /// </summary>
+    public class StringRandomizerService : IStringRandomizerService
+    {
+        //TODO - When attempting to write UTs for this service, found out I couldn't because Realm class doesn't implement an interface/virtual methods.
```
AlphaGit, line 14: I think this might be a showstopper for this seed part. If it does not allow for testing, it will become an issue to the projects generated from this seed.

AlphaGit (changes requested):

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Services/StringRandomizerService.cs
mvazquez-ms: Yep, I mentioned this on PR's description..

> One drawback is that Realm class doesn't implement an interface, so a framework other than Moq will be required for UnitTests (one capable of mockings non virtual methods, like TypeMock).

So, I won't merge it until I'm sure we can mock it with TypeMock or other framework.

cc-ing @matiasbeckerle , since he had the same concern.

mvazquez-ms:
After reviewing RealmDB we decided to disregard it as a local storage alternative.

The main reasons are:

- It's not testeable (at least not without a considerable effort)
- Due to it's design, entities are tightly coupled to Realm (doesn't suit our needs/standards)
- The overall implementation aims for simplicity (which could be good for a quick PoC), but, IMO, it was taken to an extreme 'database for dummies' implementation, were consumers have no saying on how they want to use it (e.g. waving code, it's not reliable, but it's the only way to use it)

PR #6, opened 2017-03-07
[MLP-122] Entity Framework Test

mvazquez-ms:
## Background

Still evaluating local storage alternatives. After [this ](https://github.com/makingsensetraining/mobile-pocs/pull/5) PR and some analysis, Realm doesn't seem to be a good choice (more details to be added on the Realm PR).

The other options are Sqlite and Couchbase-Lite (relatively new). So I wanted to try Sqlite, I would like to use a lightweight ORM like Dapper, but since it relies on Reflection.Emit, it's not compatible with iOS's AOT compilation.

Somehow, I ended up trying Entity Framework (not my favourite ORM), but seems to be working fine.

The other option is to do something similar to what we have on DAL2, a simple API to store/retrieve serialized objects (to avoid handling schema, migrations, etc.. but now without drawbacks).

So this PR is to gather early feedback and opinions about using EF on Xamarin. Again, the code is simple and ugly at times, but it's just to test that it works and it's a viable option.

## Changes done

- Referenced Entity Framework
- Created a sample Entity and LocalStorage DbContext
- Created a console project to be able to generate migrations (EF tools are not available yet on other platforms)

## Notes

If we move forward with this approach, we'll have an interface for the DbContext and all those nice things we like to do on WebAPI projects.


matiasbeckerle:
Before starting the review: I love code simplicity of EF but... is not overkill for mobile apps? I will do a quick search for more info on the matter.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/packages.config
```diff
@@ -2,30 +2,51 @@
 <packages>
   <package id="Autofac" version="4.4.0" targetFramework="monoandroid71" />
   <package id="Autofac.Extras.MvvmCross" version="4.0.2" targetFramework="monoandroid71" />
+  <package id="Microsoft.CSharp" version="4.3.0" targetFramework="monoandroid71" />
```
matiasbeckerle, line 5: Why this? Just being curious.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/LocalStorageContext.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.EntityFrameworkCore;
+
+namespace MvvmSeed.Domain.Model
+{
+    public class LocalStorageContext : DbContext
+    {
+        public DbSet<RandomizedString> RandomizedStrings { get; set; }
+
+        private string DatabasePath { get; set; }
```
matiasbeckerle: So I guess you are going to assume local storage is always a database instead of a file, memory or whatever... right? Asking because the names you are using. I remember when you proposed me to have something agnostic instead.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/Migrations/20170307130524_Initial.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace MvvmSeed.Domain.Model.Migrations
+{
+    public partial class Initial : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateTable(
```
matiasbeckerle, line 10: This was auto-generated based on model?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/RandomizedString.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace MvvmSeed.Domain.Model
+{
+    public class RandomizedString
+    {
+        public const long DefaultRandomizedStringPrimaryKey = 1;
```
matiasbeckerle: Are you using this?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Services/StringRandomizerService.cs
```diff
@@ -0,0 +1,44 @@
...
+            var randomizedString = new string(input.ToCharArray().OrderBy(s => random.Next(2) % 2 == 0).ToArray());
+
+            var lastRandomizedString = GetLastEntyFromDb() ?? new RandomizedString();
+            lastRandomizedString.LastTransformationTimestamp = DateTimeOffset.UtcNow;
+            lastRandomizedString.SampleTransformationCount++;
+            lastRandomizedString.LastTransformationValue = randomizedString;
+            if (lastRandomizedString.Id == 0)
+                _dbContext.Add(lastRandomizedString);
+
+            _dbContext.SaveChanges();
```
matiasbeckerle: Much better than previous code!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/Migrations/LocalStorageContextModelSnapshot.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+
+namespace MvvmSeed.Domain.Model.Migrations
+{
+    [DbContext(typeof(LocalStorageContext))]
+    partial class LocalStorageContextModelSnapshot : ModelSnapshot
```
matiasbeckerle, line 8: What is this?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/LocalStorageContext.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.EntityFrameworkCore;
+
+namespace MvvmSeed.Domain.Model
+{
+    public class LocalStorageContext : DbContext
+    {
+        public DbSet<RandomizedString> RandomizedStrings { get; set; }
```
matiasbeckerle: Minor: this one should be below the constructor.

matiasbeckerle:
Well... they say this version (the core one) is lightweight and multi platform. It seems the way to go... only experience would tell otherwise.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.MigrationsTool/Program.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace MvvmSeed.MigrationsTool
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            //NOTES: We need this console projects to generate migrations since EF Tools are not available for the other platforms yet
+            //TODO: Do something cool here to generate migrations automatically
+            Console.WriteLine("WIP");
```
matiasbeckerle, line 11: This would be a problem? A no-go maybe?

matiasbeckerle (commented):
Awesome job. Leave a couple of questions in order to have a better understanding.

AlphaGit (approved):
Looks really good! I imagine that EF solves the testability problem that we had before, right?

Aside from that, I saw that much of the code is boilerplate for getting this to work, but the migrations, the actual code using EF and the service is really simple and I love that.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/packages.config
mvazquez-ms: Probably a dependency from the other packages.. the new .NET standard approach relies heavily on packages instead of local versions of .NET framework.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/Migrations/20170307130524_Initial.cs
mvazquez-ms: Yep.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/Migrations/LocalStorageContextModelSnapshot.cs
mvazquez-ms: Check [this](http://www.learnentityframeworkcore.com/migrations/model-snapshot).

(I don't know that much about EF to explain it myself)

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.MigrationsTool/Program.cs
mvazquez-ms: I don't think so.. Right now I'm generating migrations manually when required, I just want to add some automation wherever's possible. Any ideas are welcome.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/LocalStorageContext.cs
mvazquez-ms: Done

mvazquez-ms:
@matiasbeckerle , @AlphaGit I would appreciate you taking another look into this PR.

I've added a couple of UTs to test how the EFCore.InMemory provider works and some minor adjustments required to test on a real device.

I'll work on the initialization flow for DbContext on a separte PR (so you'll see a couple of TODO related to that)

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/Migrations/LocalStorageContextModelSnapshot.cs
AlphaGit: This seems to be the solution to several problems that we used to have with EF:

- The need for a DB to actually create a new migration (even before coding it), because EF went to the DB to check where it was to automatically add pending changes to your to-be-created migration. This also means that you can have a single DB for your whole team instead of each local database.
- The need for those ugly resource files for each migration (no wonder why your approach here looks cleaner!)

This article explains it a lot longer, but a lot better: https://msdn.microsoft.com/en-us/library/dn481501(v=vs.113).aspx

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Modules/ApplicationModule.cs
```diff
@@ -9,6 +13,28 @@ public class ApplicationModule : Module
...
+        // TODO [TechnicalDebt | WIP] Improve DbContext initialization (this method is ugly, also change it to EnsureCreatedAsync)
+        private LocalStorageContext InitializeLocalStorageContext(string dbPath)
+        {
+
+            var filePath = Path.Combine(dbPath, "localStorage.db");
+            if (!Directory.Exists(dbPath))
+                Directory.CreateDirectory(dbPath);
+            if (!File.Exists(filePath))
+            {
+                using (var stream = File.Create(filePath)) { /* DB context creation fails on real devices if the file doesn't exit, review while improving DbContext initialization*/ }
```
AlphaGit: Look into the Database Initialization options for EF. Although, since you're looking for something specific to SqlLite, it may not be exactly the same.

AlphaGit (approved):
Looks great so far!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/Migrations/LocalStorageContextModelSnapshot.cs
matiasbeckerle: Thanks!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Modules/ApplicationModule.cs
```diff
@@ -9,6 +13,28 @@ public class ApplicationModule : Module
...
+            // TODO [TechnicalDebt | WIP] Improve DbContext initialization
+            cb.Register(c => InitializeLocalStorageContext(c.ResolveNamed<string>(App.BootstrapParamaters.LocalAppDataFolder)));
+
+            cb.Register(c => new StringRandomizerService(c.Resolve<LocalStorageContext>())).As<IStringRandomizerService>().SingleInstance();
+        }
+
+        // TODO [TechnicalDebt | WIP] Improve DbContext initialization (this method is ugly, also change it to EnsureCreatedAsync)
+        private LocalStorageContext InitializeLocalStorageContext(string dbPath)
+        {
+
```
matiasbeckerle: Minor: unnecessary blank line.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Modules/ApplicationModule.cs
matiasbeckerle: Typo: missing *s* in "...doesn't exit..."

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Services/StringRandomizerService.cs
```diff
@@ -0,0 +1,44 @@
...
+using MvvmSeed.Domain.Model;
+
+namespace MvvmSeed.Domain.Services
+{
+    /// <summary>
+    /// Simple implementation of <see cref="IStringRandomizerService"/> that stores randomization info on a local database (Realm)
+    /// </summary>
+    public class StringRandomizerService : IStringRandomizerService
+    {
+        private readonly LocalStorageContext _dbContext;
```
matiasbeckerle: As I said in my first review: there are different vocabulary applying in these scenarios. I think you should standardize it because it's kind of difficult for reading through all the classes (at least for me).

Look at this class for example. `LocalStorageContext` - `_dbContext` - `local database`.

I think it will read better if you use always the same wording. Also: this is a local storage approach where you don't care about being a db, file or remote service? If that is the case I believe you should not use the db wording.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Services/StringRandomizerService.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Linq;
+using MvvmSeed.Domain.Model;
+
+namespace MvvmSeed.Domain.Services
+{
+    /// <summary>
+    /// Simple implementation of <see cref="IStringRandomizerService"/> that stores randomization info on a local database (Realm)
```
matiasbeckerle: I think _(Realm)_ doesn't apply here.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Services/StringRandomizerService.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public DateTimeOffset LastRandomizationTimestamp => GetLastEntyFromDb()?.LastTransformationTimestamp ??  DateTimeOffset.MinValue;
+
+        public int RandomizationsCount => GetLastEntyFromDb()?.RandomizationCount ?? 0;
+
+        public string LastRandomizedValue => GetLastEntyFromDb()?.LastTransformationValue ?? "Hello World!";
+
+        private RandomizedString GetLastEntyFromDb() => _dbContext.RandomizedStrings.FirstOrDefault();
+
```
matiasbeckerle: Minor: unnecessary blank line.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.MigrationsTool/Program.cs
matiasbeckerle: So creating migrations is possible on other platforms but the only issue is that the process is manual?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.UnitTests/Services/StringRandomizerServiceTest.cs
```diff
@@ -0,0 +1,152 @@
+﻿using System;
+using System.Runtime.CompilerServices;
+using Microsoft.EntityFrameworkCore;
+using MvvmSeed.Domain.Model;
+using MvvmSeed.Domain.Services;
+using Xunit;
+
+namespace MvvmSeed.UnitTests.Services
+{
+    public class StringRandomizerServiceTest
```
matiasbeckerle, line 10: Really nice these ones...

matiasbeckerle (approved):

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.MigrationsTool/Program.cs
mvazquez-ms: By manualIly I meant opening a console and running commands using `dotnet ef migrations`.

The problem is `EntityFrameworkCore.Tools.DotNet` cannot be installed on a .NET Standard project (yet). I would like to review this process to minimize the steps or at least have a detailes document on how to do it.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Services/StringRandomizerService.cs
mvazquez-ms: Good catch!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Services/StringRandomizerService.cs
mvazquez-ms: > Also: this is a local storage approach where you don't care about being a db, file or remote service?

I'm not sure yet, but chances are this design will assume you have a database, at least initially. I want to keep complexity low.

As for the names, I'll see what can I do there.

matiasbeckerle: Thanks!

PR #7, opened 2017-03-10
[MLP-122] Improvements on app initialization

mvazquez-ms:
## Background

Second part to [this ](https://github.com/makingsensetraining/mobile-pocs/pull/6) PR.

On this one I deal with app initialization.

## Changes done

- Added a SplashView (setup theme, layout, background initialization, etc)
- Created an App Initializer to be executed while splash view is shown (initializes DbContext)
- Added `LinkerPleaseInclude` (comments on the class explaining why)
- Created a `DbContextBase` class to simplify `LocalStorageContext`


Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Linker/LinkerPleaseInclude.cs
```diff
@@ -0,0 +1,85 @@
...
+using Android.Views;
+using Android.Widget;
+
+namespace MvvmSeed.Android.Linker
+{
+    /// <summary>
+    /// This file is required to avoid android Linker wiping references to code not detected during static analysis.
+    /// (We could disable linking, but I rather not, no linking means x3 or x4 size for builds)
+    /// </summary>
+    public class LinkerPleaseInclude
```
matiasbeckerle, line 13: This class is something that will grown over time... right? If that is the case I think it will be helpful to clarify in the comment above which classes should be added and why.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/MvvmSeed.Android.csproj
```diff
@@ -138,10 +138,12 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="Linker\LinkerPleaseInclude.cs" />
     <Compile Include="Resources\Resource.Designer.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Setup.cs" />
     <Compile Include="Views\SampleView.cs" />
+    <Compile Include="Views\SplashView.cs" />
```
matiasbeckerle, line 146: Really nice that you included this.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Resources/layout/View_Splash.axml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+              android:orientation="vertical"
+              android:layout_width="fill_parent"
+              android:layout_height="fill_parent">
+  <TextView
+    android:layout_width="fill_parent"
+    android:layout_height="wrap_content"
+    android:text="Loading...." />
```
matiasbeckerle: Really minor but... do you mind using only 3 dots instead of 4?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
```diff
@@ -19,19 +27,29 @@ protected override IMvxIoCProvider CreateIocProvider()
...
-
-            // TODO [TechnicalDebt | WIP] Improve DbContext initialization
-            // Set platform-specific path to local app data folder
-            containerBuilder.Register(c => System.Environment.GetFolderPath(System.Environment.SpecialFolder.Personal)).Named<string>(App.BootstrapParamaters.LocalAppDataFolder);
+        protected override void InitializeViewLookup()
+        {
+            // ViewModel->View relations
+            // Used by MvvmCross framework to automatically resolve ShowViewModel<TViewModel> into a view
+            // MvvmCross by default uses naming convention, but this approach provides flexibility (we can still rely on naming conventions if we want to and override only when required)
+            var viewModelToViewMapping = new Dictionary<Type, Type>
```
matiasbeckerle, line 39: Minor but it looks like this could grow up. What do you think of using a separated file to only put this VM to V mapping? Or something similar.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Views/SplashView.cs
```diff
@@ -0,0 +1,19 @@
+using Android.App;
+using Android.Content.PM;
+using MvvmCross.Core.ViewModels;
+using MvvmCross.Droid.Views;
+using MvvmCross.Platform;
+
+namespace MvvmSeed.Android.Views
+{
+    [Activity(MainLauncher = true, NoHistory = true, ScreenOrientation = ScreenOrientation.Portrait, Theme = "@style/Theme.Splash")]
```
matiasbeckerle, line 9: Why portrait? What happens if I'm using the phone on landscape and I open this app?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Modules/ApplicationModule.cs
```diff
@@ -12,29 +12,16 @@ public class ApplicationModule : Module
     {
         protected override void Load(ContainerBuilder cb)
         {
-            cb.Register(c => new MvxAppStart<SampleViewModel>()).As<IMvxAppStart>().SingleInstance();
-
-            // TODO [TechnicalDebt | WIP] Improve DbContext initialization
-            cb.Register(c => InitializeLocalStorageContext(c.ResolveNamed<string>(App.BootstrapParamaters.LocalAppDataFolder)));
+            cb.Register(c => c.Resolve<SplashViewModel>()).As<IMvxAppStart>().SingleInstance();
```
matiasbeckerle, line 15: God this class is looking amazing. Keep it without the dirty hacks 😄 !

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/ViewModels/SplashViewModel.cs
```diff
@@ -0,0 +1,20 @@
...
+        private readonly IApplicationInitializer _appInitializer;
+
+        public SplashViewModel(IApplicationInitializer appInitializer)
+        {
+            _appInitializer = appInitializer;
+        }
+
+        public void Start(object hint = null)
+        {
+            _appInitializer.PostBootsrapInitializationAsync().ContinueWith(cf => ShowViewModel<SampleViewModel>());
```
matiasbeckerle, line 17: Nice. An alternative of using events.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/DbContextBase.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System.IO;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+
+namespace MvvmSeed.Domain.Model
+{
+    public abstract class DbContextBase : DbContext
+    {
+        private enum DatabaseMode
```
matiasbeckerle, line 9: Keimeil will be glad to see this enum in another file.

matiasbeckerle: Unless you want to keep it private for a reason.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/LocalStorageContext.cs
```diff
@@ -2,23 +2,22 @@
...
 namespace MvvmSeed.Domain.Model
 {
-    public class LocalStorageContext : DbContext
+    public class LocalStorageContext : DbContextBase
     {
-        private readonly string _databasePath;
+        private const string DatabaseName = "LocalStorage.db";
 
+        /// <summary>
+        /// Initializes a InMemory database context to be used on UnitTests
```
matiasbeckerle: Minor: I think it should stand `an InMemory`.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/LocalStorageContext.cs
```diff
@@ -2,23 +2,22 @@
...
 
-        public LocalStorageContext(string databasePath)
-        {
-            _databasePath = databasePath;
-        }
+        /// <summary>
+        /// Initializes a Sqlite database context
+        /// </summary>
+        /// <param name="databaseFolder">Full path where the database will be created/open </param>
+        public LocalStorageContext(string databaseFolder) : base(databaseFolder, DatabaseName) { }
```
matiasbeckerle, line 19: Maybe db name could be provided as param?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/DbContextBase.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System.IO;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+
+namespace MvvmSeed.Domain.Model
+{
+    public abstract class DbContextBase : DbContext
```
matiasbeckerle, line 7: Something is bugging me about the responsibility here that I'm not sure if is right or wrong. This class looks like a base, something generic but it also has some sqlite-specific code. Again, not sure... but you are the single responsibility principle dude.

matiasbeckerle (commented):
Besides all the questions and comments I put here you should consider printing this PR and put it on a frame. Code reads really easy and overall these changes kickass.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Linker/LinkerPleaseInclude.cs
mvazquez-ms: It's hard to know which classes should be added and why 😄 

Most of the references added now are needed because MvvmCross' binding mechanisms works through reflection and a static code analysis won't catch those.

I'm not sure I can be more _precise_ than 'code not detected during static analysis'. I think this file will grow once we start seeing `MissingMethodExceptions` on release builds, not something we can easily anticipate. But it's not something that will happen often and most frameworks indicate if they need extra linker configuration for release builds.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Resources/layout/View_Splash.axml
mvazquez-ms: Sure!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
mvazquez-ms: Yep, if it ends up growing, we should definitely move it to a separate file/method. But I prefer to see how this evolves first, I'm not 100% we really want to do this, rely on naming conventions or use some custom attribute.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Views/SplashView.cs
mvazquez-ms: Your phone will implode.. 

Or, the splash view will be shown in portrait mode regardless your phone orientation. Usually splash views are portrait-only.. if we want orientation-aware splash, we'll need a different asset for the background, and in that case we remove that attribute and adjust the theme accordingly.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/DbContextBase.cs
mvazquez-ms: Yep, I thought the same while coding this.. 

> you are the single responsibility principle dude.

Hehe, yep, but I'm also the KISS principle dude when the degree of certainty is not enough to decide what approach to take. In such case, I prefer to write simple code and let the class evolve naturally.

The thing that bothers me most of this class is that I'm taking two different approaches for the same problem. Both Sqlite and InMemory are separate packages and added as extension methods to `DbContextOptionsBuilder`. I'm assuming Sqlite usage on Domain and letting UT project handle InMemory mode by itself.

Since the code is extremely simple, it's not a big deal right now, but I'll probably decide for one or the other soon, in a separate PR.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/DbContextBase.cs
mvazquez-ms: Yep, I want to keep it private for now, but related to the comment above, I might remove that enum completely.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/LocalStorageContext.cs
mvazquez-ms: Yep, I prefer to hide that from consumers.. could be worth it on some scenarios, but could be error-prone too. So, again, I'll let it evolve naturally, if the need arises, we'll add the parameter for db name.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/LocalStorageContext.cs
mvazquez-ms: Fixed!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Linker/LinkerPleaseInclude.cs
AlphaGit: What if we specified manually the bindings? It would also give us finer control over the bindings. Yes, we would end up with a huge file that has access to everything, which is a downside, but it prevents us from having to do any of this and it also increases visibility on the configuration of the bindings.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
```diff
@@ -19,19 +27,29 @@ protected override IMvxIoCProvider CreateIocProvider()
         {
             var containerBuilder = new ContainerBuilder();
             containerBuilder.RegisterModule<ApplicationModule>();
+            containerBuilder.Register(c => Environment.GetFolderPath(Environment.SpecialFolder.Personal)).Named<string>(App.BootstrapParamaters.LocalAppDataFolder);
```
AlphaGit, line 30: Is this parameter the name given to the registration? If so, maybe we should go with a constant value?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Views/SplashView.cs
AlphaGit: > Your phone will implode..

That's a feature that has been requested for ages. Yay!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Modules/ApplicationModule.cs
```diff
@@ -12,29 +12,16 @@ public class ApplicationModule : Module
...
+            cb.Register(c => new ApplicationInitializer(c.Resolve<LocalStorageContext>())).As<IApplicationInitializer>();
 
+            cb.Register(c => new LocalStorageContext(c.ResolveNamed<string>(App.BootstrapParamaters.LocalAppDataFolder))).As<LocalStorageContext>().SingleInstance();
```
AlphaGit, line 18: Same question here about the registration name.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/DbContextBase.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <summary>
+        /// Initializes database file
+        /// </summary>
+        public async Task InitializeAsync()
+        {
+            if (_databaseMode != DatabaseMode.Sqlite)
+                return;
+
+            // TODO - Double check the need for this code after the fix for Android 7 is released (https://github.com/aspnet/EntityFramework/issues/7777)
+            await Task.Run(() =>
```
AlphaGit, line 64: Do you need `await Task.Run()` at all?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/DbContextBase.cs
```diff
@@ -0,0 +1,74 @@
...
+            {
+                if (!Directory.Exists(_databaseFolder))
+                    Directory.CreateDirectory(_databaseFolder);
```
AlphaGit, line 67: Are there async equivalent of these methods? Maybe you can await on those.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Linker/LinkerPleaseInclude.cs
matiasbeckerle: I think the explanation you added on the latest comment is something useful to add it. Specially for new team members. At least to have an idea of what it is. Maybe is something really common in Android universe, but at least for me is weird.

mvazquez-ms: > What if we specified manually the bindings?

I'm not sure if I follow..

Right now bindings are defined on View's axml like this:

```
  <TextView
    android:layout_width="fill_parent"
    android:layout_height="wrap_content"
    local:MvxBind="Text SampleString"/>
```

You are proposing writing those bindings on View's code-behind? I'm not a big fan of that. The linker-stuff is ugly too, but it's not something that will require a lot of maintenance, it's not like we'll have to add code to that file often.. Most frameworks are aware of this and provides support for linker issues (e.g. [linker issue](https://github.com/MvvmCross/MvvmCross/issues/689) ).

I'm not saying no, I'm saying that writing bindings on code behind it's a heavy decision, if that's what you had in mind.

> Yes, we would end up with a huge file that has access to everything

That's pretty much what this file does 😄 . The linker is wiping references to `button.Click` for example, because there's not a single line on the code that references button's Click event. MvvmCross does use it internally through commands.. So that huge file will end up being just dummy reference to pieces of code that are not directly referenced by our own code, which is what this file is.

So, maybe I'm not understanding your proposal @AlphaGit

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
mvazquez-ms: So we have:

`containerBuilder.Register(c => {value}).Named<string>({name});`

I'm registering the value `Environment.GetFolderPath(Environment.SpecialFolder.Personal)`, and tagging it with the name `App.BootstrapParamaters.LocalAppDataFolder`

So, I'm saying `LocalAppDataFolder` path for android is `Environment.GetFolderPath(Environment.SpecialFolder.Personal)` value.

Then I'm using that value when creating a `LocalStorageContext` referencing it with

`c.ResolveNamed<string>(name: App.BootstrapParamaters.LocalAppDataFolder)`

>Is this parameter the name given to the registration? If so, maybe we should go with a constant value?

The name for the parameter is a constant value. Are you seeing something weird on this code?

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/DbContextBase.cs
mvazquez-ms: I think so.. I need that before calling `EnsureCreatedAsync()` but I want to do I/O stuff on a separate thread.

`File` and `Directory` classes on .NET doesn't provide async support, maybe there are alternatives for those..

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/DbContextBase.cs
mvazquez-ms: Not that I know of, but I'll check again.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
matiasbeckerle: Seems fair, thanks!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Linker/LinkerPleaseInclude.cs
AlphaGit: Nope, I think you got it right.

I am also not a big fan of having a huge code behind with bindings, I definitely think XAML is the place for that.

Maybe we can find / write something that will link the right classes by looking at the XAML bindings?

What I'd really like to avoid is someone needing to remember to add a line into this class because they forgot before and the released version of the app does not work. The more automatic, the better.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Views/SplashView.cs
matiasbeckerle: LOL Actually I think Samsung already did that... @AlphaGit https://www.wired.com/2017/01/why-the-samsung-galaxy-note-7-kept-exploding/

matiasbeckerle: Besides jokes, I saw the behavior you mention @mvazquez-ms. Maybe is fine for the consumers but it will be great to have something that works for both scenarios. No big deal anyway.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/DbContextBase.cs
matiasbeckerle: Awesome that you feel the same and keep it simple for now. Seems fair for me!

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/LocalStorageContext.cs
matiasbeckerle: Fair enough.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
AlphaGit: Why using the value from `App.BootstrapParamaters.LocalAppDataFolder` (which could be, `/home/someUser/AppData`) instead of using just the value `"localAppData"`. This is a registration, so the actual name of the registration does not need to be the actual contents. (In fact, I think it shouldn't).

So, right now you'll have in your registrations table:

Registration Name: `/home/someUser/AppData`, Value: `/home/someUser/Personal`

I suggest we should have:

Registration Name: `ApplicationData`, Value: `/home/someUser/Personal`

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Domain/Model/DbContextBase.cs
AlphaGit: Ah, gotcha.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Android/Setup.cs
mvazquez-ms: Maybe I didn't explained it right..

>Why using the value from App.BootstrapParamaters.LocalAppDataFolder (which could be, /home/someUser/AppData)

That's a constant and I'm using it as parameter name. That const is defined at application level (platform-independent), as:

```
public static class BootstrapParamaters
{
      public const string LocalAppDataFolder = "LocalAppDataFolder";
}
```

> so the actual name of the registration does not need to be the actual contents. (In fact, I think it shouldn't).

Exactly, it's not.

> So, right now you'll have in your registrations table:
> Registration Name: /home/someUser/AppData, Value: /home/someUser/Personal
> I suggest we should have:
> Registration Name: ApplicationData, Value: /home/someUser/Personal

What you're suggesting is what we have right now.

Maybe the confusion came because that const was defined on the other PR. Do you think we can make it more obvious if we rename something?

AlphaGit: > That const is defined at application level (platform-independent), as:

Ahhhhhhhhhh got it. Sorry, it looked so familiar that I thought we were actually reading the value from the environment. My bad.

Xamarin/MvvmSeed/MvvmSeed/MvvmSeed.Application/Modules/ApplicationModule.cs
AlphaGit: Disregard this comment.

AlphaGit (approved):

PR #10, opened 2017-03-22
[MLP-150] First implementation of Core Data

jsmuguruza:
Hey guys! this is just the first implementation of Core Data, it's quite basic but it's the first time I implemented it by myself (used other devs implementation in the past). I had some issues while implementing it so I ended up with only write functionality as of now and just one entity, quite simple. My idea for tomorrow it's to add read and modify functionality, and maybe some relations between entities, I don't know.

The idea of this PR it's for you just to give some feedback in how thing are being implemented ATM, and I also left some comments in the code on purpose, so feel free to leave as much comments as you want so I can improve it tomorrow.

Kudos!

### DIFF
--- a/IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
+++ b/IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
@@ -28,34 +28,5 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func applicationDidBecomeActive(_ application: UIApplication) { }
 
     func applicationWillTerminate(_ application: UIApplication) { }
-    
-    // MARK: - Core Data stack
-    //TODO: Evaluate if having Core Data's stack in AppDelegate is the best approach.
-    lazy var persistentContainer: NSPersistentContainer = {
-        let container = NSPersistentContainer(name: "SwiftSeedModel")
-        container.loadPersistentStores(completionHandler: { (storeDescription, error) in
-            if let error = error as NSError? {
-                // TODO: Proper Error handling should be done here.
-                // fatalError() causes the application to generate a crash log and terminate. While useful in development this should be changed later. - Mugu.
-                fatalError("Unresolved error \(error), \(error.userInfo)")
-            }
-        })
-        
-        return container
-    }()
-    
-    // MARK: - Core Data Saving support
-    func saveContext () {
-        let context = persistentContainer.viewContext
-        if context.hasChanges {
-            do {
-                try context.save()
-            } catch {
-                // TODO: Proper Error handling should be done here.
-                let nserror = error as NSError
-                fatalError("Unresolved error \(nserror), \(nserror.userInfo)")
-            }
-        }
-    }
 }
 
--- a/IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
+++ b/IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
@@ -7,6 +7,7 @@
 //
 
 import UIKit
+import CoreData
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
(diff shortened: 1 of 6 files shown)


mmaldini33 (approved):

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
```diff
@@ -18,26 +19,43 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    func applicationDidEnterBackground(_ application: UIApplication) { }
+
+    func applicationWillEnterForeground(_ application: UIApplication) { }
+
+    func applicationDidBecomeActive(_ application: UIApplication) { }
+
+    func applicationWillTerminate(_ application: UIApplication) { }
+    
+    // MARK: - Core Data stack
+    //TODO: Evaluate if having Core Data's stack in AppDelegate is the best approach.
```
mvazquez-ms: I think we should move core data related code to another class, it's not app delegate responsibility.

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
```diff
@@ -18,26 +19,43 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    lazy var persistentContainer: NSPersistentContainer = {
```
mvazquez-ms: What are the reasons to make this lazy?

The problem I see with this is that storage initialization could fail at any time, and usually it's better to check/initialize dependencies at startup, and be able to recover/inform that error as soon as possible.

Lazy could lead to an scenario where the error is thrown inside a ViewController that has nothing to do with CoreData initialization.

mvazquez-ms: Also, this shouldn't be internal, that's the same as having global access to database.

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
```diff
@@ -18,26 +19,43 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        let container = NSPersistentContainer(name: "SwiftSeedModel")
+        container.loadPersistentStores(completionHandler: { (storeDescription, error) in
+            if let error = error as NSError? {
```
mvazquez-ms: How is error defined? Could it be some other type than `NSError` ? If not, can we avoid the cast? If yes, can we log an error on that case?

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
```diff
@@ -18,26 +19,43 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                // TODO: Proper Error handling should be done here.
```
mvazquez-ms: Agree, another good topic to discuss on this seed, proper error handling!

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
```diff
@@ -18,26 +19,43 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                // fatalError() causes the application to generate a crash log and terminate. While useful in development this should be changed later. - Mugu.
+                fatalError("Unresolved error \(error), \(error.userInfo)")
+            }
+        })
+        
+        return container
+    }()
+    
+    // MARK: - Core Data Saving support
+    func saveContext () {
```
mvazquez-ms: Same here, default accessibility modifier is not restrictive enough for a method like `saveContext`

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
```diff
@@ -18,26 +19,43 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        let context = persistentContainer.viewContext
+        if context.hasChanges {
+            do {
+                try context.save()
+            } catch {
+                // TODO: Proper Error handling should be done here.
+                let nserror = error as NSError
```
mvazquez-ms: Where does `error` come from? 😮 Is it something specific to `catch` blocks?

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
mvazquez-ms: Minor, whitespace between method name and parentheses.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/AddPersonViewController.swift
```diff
@@ -0,0 +1,28 @@
...
+class AddPersonViewController: UIViewController {
+
+    @IBOutlet weak var txtFieldName: UITextField!
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+    }
+
+    @IBAction func btnAddPersonTouchUpInsided(_ sender: Any) {
+        let context = (UIApplication.shared.delegate as! AppDelegate).persistentContainer.viewContext //Not sure if there's a better way to handle context rather than this way.
```
mvazquez-ms: > //Not sure if there's a better way to handle context rather than this way.
  
DI 😄 (We should integrate Typhoon)

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/AddPersonViewController.swift
```diff
@@ -0,0 +1,28 @@
...
+
+class AddPersonViewController: UIViewController {
+
+    @IBOutlet weak var txtFieldName: UITextField!
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+    }
+
+    @IBAction func btnAddPersonTouchUpInsided(_ sender: Any) {
```
mvazquez-ms, line 20: Even if we have a MVC approach right now, controllers shouldn't be responsible of handling CoreData's context.

We should have a `PersonService` and this controller should delegate to it.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/AddPersonViewController.swift
mvazquez-ms: This is related to the lazy comment.. being `persistentContainer` lazy, clicking on AddPerson button could fail, and there's not way to recover from the error at this point.

mvazquez-ms (changes requested):
I know this was just a simple example and some of my comments might not be solved on this PR, just consider them as guidelines on how this code should evolve.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/AddPersonViewController.swift
mvazquez-ms: Also, there's the performance hit of `persistentContainer` initialization.. and it's better to hide that on splash/loading screen instead of a user interaction like this button.

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
jsmuguruza: Right, I figured that you would like to approach it that way, still a lot of people use it on AppDelegate (not saying it's a good idea, just pointing that out). Nevertheless, I also saw that a lot of people also use it on a Singleton. What do you think about it, how could we avoid using a singleton but still having a class that's accesible when we need it?

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
jsmuguruza: Well, I was following some tutorials and documentation and I saw that most of them use lazy initialization, my thoughts on it are that, as the `loadPersistentStores` uses a block and completion handler, using lazy make it only be usable when it's finally initialized, but then again, I don't have that much experience using lazy and I might be wrong about it. 

But I guess you're right about the error being thrown inside a ViewController, or any other class. I will see how can I handle this, thanks!

jsmuguruza: Right.

lpelizzaMS (changes requested):
I agree with the comments that Marcelo has made

PR #14, opened 2017-03-30
[MLP-150] - Implement core data stack

lpelizzaMS:
# Background

We need to implement the logic to persist data on Swift. Our first approach is Core Data.

# Changes done
Create the core data stack. 
Create a  data model
Added some lines of code on AppDelegate to test the stack

# Approvers required
@jsmuguruza | @mvazquez-ms | @mmaldini33 | @bsztamfater-ms

### DIFF
--- a/IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
+++ b/IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
@@ -14,17 +14,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
     private let container = DependencyContainer.configure()
-    
+
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         BuddyBuildSDK.setup()
         
-        // Override point for customization after application launch.
         return true
     }
 
     func applicationWillResignActive(_ application: UIApplication) {
-        // Sent when the application is about to move from active to inactive state. This can occur for certain types of temporary interruptions (such as an incoming phone call or SMS message) or when the user quits the application and it begins the transition to the background state.
-        // Use this method to pause ongoing tasks, disable timers, and invalidate graphics rendering callbacks. Games should use this method to pause the game.
+        let coreDataStack = try! container.resolve() as CoreDataStack
+        coreDataStack.saveChanges()
     }
 
     func applicationDidEnterBackground(_ application: UIApplication) {
@@ -41,7 +40,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func applicationWillTerminate(_ application: UIApplication) {
-        // Called when the application is about to terminate. Save data if appropriate. See also applicationDidEnterBackground:.
+        let coreDataStack = try! container.resolve() as CoreDataStack
+        coreDataStack.saveChanges()
     }
 }
 
--- a/IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/UserService.swift
+++ b/IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/UserService.swift
@@ -1,0 +1,33 @@
+//
+//  UserService.swift
+//  SwiftSeedProject
+//
+//  Created by Lucas Pelizza on 3/31/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+
+final class UserService {
+    
+    private let persistence: Persistence
+    
+    //TODO: this need to be changed. I need to think more about it and fix the PersistenceObject
+    private let entityName = "User"
+    
+    init(persistence: Persistence) {
+        self.persistence = persistence
+    }
+    
+    public func getUsers() -> [User] {
+        return persistence.getAll(entityName: entityName)
+    }
+    
+    public func getUserBy(identifier: Int) -> User {
+        return persistence.getBy(entityIdentifier: identifier, entityName: entityName)
+    }
+    
+    public func add(user: User) {
+        persistence.add(entity: user, entityName: entityName)
+    }
+}
--- a/IOS/SwiftSeedProject/SwiftSeedProject/Domain/User+CoreDataProperties.swift
+++ b/IOS/SwiftSeedProject/SwiftSeedProject/Domain/User+CoreDataProperties.swift
@@ -1,0 +1,24 @@
+//
+//  User+CoreDataProperties.swift
+//  SwiftSeedProject
+//
+//  Created by Lucas Pelizza on 4/10/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+
+extension User {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<User> {
+        return NSFetchRequest<User>(entityName: "User")
+    }
+
+    @NSManaged public var firstName: String?
+    @NSManaged public var id: Int16
+    @NSManaged public var lastName: String?
+    @NSManaged public var username: String?
+
+}
--- a/IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/CoreDataStack/CoreDataStack.swift
+++ b/IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/CoreDataStack/CoreDataStack.swift
@@ -1,0 +1,115 @@
+//
+//  CoreDataStack.swift
+//  SwiftSeedProject
+//
+//  Created by Lucas Pelizza on 3/30/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+class CoreDataStack {
+    
+    typealias CoreDataStackCompletion = () -> ()
+    
+    private let modelName: String
+    private let completion: CoreDataStackCompletion
+    
+    private var persistentStoreURL: NSURL {
+        let storeName = "\(modelName).sqlite"
+        let fileManager = FileManager.default
+        let documentsDirectoryURL = fileManager.urls(for: .documentDirectory, in: .userDomainMask)[0]
+        return documentsDirectoryURL.appendingPathComponent(storeName) as NSURL
+    }
+    
+    private lazy var managedObjectModel: NSManagedObjectModel? = {
+        // The mom and momd files are compiled versions of xcdatamodel and xcdatamodeld files.
+        guard let modelURL = Bundle.main.url(forResource: self.modelName, withExtension: "momd") else {
+            return nil
+        }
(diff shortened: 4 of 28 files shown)


jsmuguruza (approved):

IOS/SwiftSeedProject/SwiftSeedProject/CoreDataManager.swift
```diff
@@ -0,0 +1,132 @@
...
+    
+    typealias CoreDataManagerCompletion = () -> ()
+    private let modelName: String
+    fileprivate let completion: CoreDataManagerCompletion
+    
+    private var persistentStoreURL: NSURL {
+
+        let storeName = "\(modelName).sqlite"
+        let fileManager = FileManager.default
+
```
mmaldini33: The only thing that I see, is that there are many blank lines within each method.

mmaldini33 (approved):

mvazquez-ms:
Can you properly rebase this branch and remove unrelated commits? It is including changes from other tasks.

@lpelizzaMS

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
```diff
@@ -13,16 +13,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
 
+    //TODO: we need to define a place to add this kind of configuration like "DataModel"
```
mvazquez-ms: Good job leaving useful comments!

IOS/SwiftSeedProject/SwiftSeedProject/CoreDataManager.swift
```diff
@@ -0,0 +1,132 @@
...
+        let fileManager = FileManager.default
+        
+        let documentsDirectoryURL = fileManager.urls(for: .documentDirectory, in: .userDomainMask)[0]
+        
+        return documentsDirectoryURL.appendingPathComponent(storeName) as NSURL
+    }
+    
+    private lazy var managedObjectModel: NSManagedObjectModel? = {
+
+        guard let modelURL = Bundle.main.url(forResource: self.modelName, withExtension: "momd") else {
```
mvazquez-ms: What does "momd" mean?

Can you add a comment or replace it with a constant with a meaningful name?

IOS/SwiftSeedProject/SwiftSeedProject/CoreDataManager.swift
```diff
@@ -0,0 +1,132 @@
...
+            return nil
+        }
+        
+        let managedObjectModel = NSManagedObjectModel(contentsOf: modelURL)
```
mvazquez-ms: Are you planning on adding code in-between these two lines? Otherwise, I'd just return the object instead of assigning it and then returning it.

IOS/SwiftSeedProject/SwiftSeedProject/CoreDataManager.swift
```diff
@@ -0,0 +1,132 @@
...
+        
+        return managedObjectModel
+    }()
+    
+    private lazy var persistentStoreCoordinator: NSPersistentStoreCoordinator? = {
+        guard let managedObjectModel = self.managedObjectModel else {
+            return nil
+        }
+        
+        let persistentStoreCoordinator = NSPersistentStoreCoordinator(managedObjectModel: managedObjectModel)
```
mvazquez-ms: Are you planning on adding code in-between these two lines? Otherwise, I'd just return the object instead of assigning it and then returning it.

IOS/SwiftSeedProject/SwiftSeedProject/CoreDataManager.swift
```diff
@@ -0,0 +1,132 @@
...
+    
+    private lazy var privateManagedObjectContext: NSManagedObjectContext = {
+        
+        let managedObjectContext = NSManagedObjectContext(concurrencyType: .privateQueueConcurrencyType)
+        managedObjectContext.persistentStoreCoordinator = self.persistentStoreCoordinator
+        
+        return managedObjectContext
+    }()
+    
+    public private(set) lazy var mainManagedObjectContext: NSManagedObjectContext = {
```
mvazquez-ms: Good job adding that private set! It's important to use proper access modifiers.

IOS/SwiftSeedProject/SwiftSeedProject/CoreDataManager.swift
```diff
@@ -0,0 +1,132 @@
...
+        setupCoreDataStack()
+    }
+    
+    private func setupCoreDataStack() {
+        let _ = mainManagedObjectContext.persistentStoreCoordinator
+        
+        DispatchQueue.global().async {
+            self.addPersistentStore()
+
+            DispatchQueue.main.async { self.completion() }
```
mvazquez-ms: I would like to hear your opinion/thoughts on this:

1.
```
DispatchQueue.global().async {
    self.addPersistentStore()
    DispatchQueue.main.async { self.completion() }
}
```

2.
```
DispatchQueue.global().async {
    self.addPersistentStore()
    self.completion()
}
```

What are the reasons to choose one or the other? And what is your preference?

cc @lpelizzaMS  @bsztamfater-ms  @jsmuguruza  @mmaldini33

mvazquez-ms (commented):
Left a couple minor comments, good job!

Just be careful with git and try to keep your branch clean, this one contains a lot of unrelated changes.

IOS/SwiftSeedProject/SwiftSeedProject/CoreDataManager.swift
lpelizzaMS: The "mom" and "momd" files are compiled versions of "xcdatamodel" and "xcdatamodeld" files.

IOS/SwiftSeedProject/SwiftSeedProject/CoreDataManager.swift
lpelizzaMS: Changed

IOS/SwiftSeedProject/SwiftSeedProject/CoreDataManager.swift
lpelizzaMS: Changed

IOS/SwiftSeedProject/SwiftSeedProject/CoreDataManager.swift
lpelizzaMS: first of all, I will tray to explain the idea of completion.
Completion, has been created with the idea of showing some feedback when the stack is ready to use. E.g. change viewController's background or change a property. 
Then, if we don't know how "completion" will be use we need to put it inside of main thread because it could be an UI change

briansztamfater: I agree with @lpelizzaMS, it is a common situation to update a UI element on completion, although if we use MVVM we should not update UI directly there. I think it should be executed on main thread because it is safer, so in that way, _what_ doing on completion should be responsibility of the developer.

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/UserService.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+
+final class UserService {
+    
+    private let persistence: Persistence
+    //TODO: this need to be changed. I need to think more about it and fix the PersistenceObject
+    private let entityName = "User"
```
mvazquez-ms: Yep, this is not something easy to solve, but at least for now I would place this constant within `User` entity.

IOS/SwiftSeedProject/SwiftSeedProject/Domain/User+CoreDataProperties.swift
```diff
@@ -0,0 +1,24 @@
...
+import CoreData
+
+
+extension User {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<User> {
+        return NSFetchRequest<User>(entityName: "User");
+    }
+
+    @NSManaged public var firstname: String?
```
mvazquez-ms: Really minor, but first name and last name are actually two words, so the casing should be `firstName` and `lastName`

IOS/SwiftSeedProject/SwiftSeedProject/Domain/User+CoreDataProperties.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+
+import Foundation
+import CoreData
+
+
+extension User {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<User> {
+        return NSFetchRequest<User>(entityName: "User");
```
mvazquez-ms: Here you should reference the aforementioned constant.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/CoreDataStack/CoreDataStack.swift
```diff
@@ -0,0 +1,118 @@
...
+            
+            print("Unable to Add Persistent Store")
+            print("\(addPersistentStoreError.localizedDescription)")
+        }
+    }
+    
+    // MARK: - Helper Methods
+    
+    public func saveChanges() {
+        mainManagedObjectContext.performAndWait({
```
mvazquez-ms, line 87: Have you heard of DRY principle? 😄 Can you try avoid repeating almost the same code twice inside this method?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/PersistenceManager.swift
```diff
@@ -0,0 +1,42 @@
...
+    
+    public func getAll<T: PersistenceObject>(entityName: String) -> [T] {
+        let objectContext = coreDataManager.privateChildManagedObjectContext()
+        let fetch = NSFetchRequest<NSFetchRequestResult>(entityName: entityName)
+        
+        do {
+            let fetchedObjects = try objectContext.fetch(fetch) as! [T]
+            return fetchedObjects
+        } catch {
+            fatalError("Failed to fetch employees: \(error)")
```
mvazquez-ms: What employees? 😮

mvazquez-ms (changes requested):

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/CoreDataStack/CoreDataStack.swift
lpelizzaMS: I will search about DRY principle.

IOS/SwiftSeedProject/SwiftSeedProject/Domain/User+CoreDataProperties.swift
lpelizzaMS: gotcha

PR #15, opened 2017-04-05
[MLP-155] Integrate a REST client library

briansztamfater:
## Background

We need to implement a REST solution. I made a research of the most popular clients, and found Alamofire and Siesta. I decided to implement Alamofire because its community is bigger, and therefore the ir much more documentation about it.

## Changes done

I implemented Alamofire with SwiftyJSON, and made 3 basic examples (GET, PUT and POST) with a dummy API. Also I integrated them with data binding so the UI elements updates when we get the server response.

I took this repo as the base of my implementation: https://github.com/ShengHuaWu/SimpleApiClient

## Approvers required

@mvazquez-ms @lpelizzaMS @mmaldini33 @jsmuguruza


IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/LoginViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+    // TODO: Create a service for authentication and remove references to Api
+    private let api: Api
+    private var user: User?
+    
+    var name = Observable<String?>("")
+    var email = Observable<String?>("")
+    
+    init() {
+        // TODO: We need to implement dependency injection
+        self.api = Api.init()
```
mvazquez-ms: Super minor, but you can write this 
`self.api = Api.init()`
as:
`self.api = Api()`

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/LoginViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+    }
+    
+    func getRandomUser() {
+        let randomId:UInt32 = arc4random_uniform(10) + 1; // Fake API has users with id from 1 to 10
+        api.getUser(userId: String(randomId)) { [weak self] result in
+            do {
+                guard let strongSelf = self else {
```
mvazquez-ms: Awesome! Complying with our style guide!

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/LoginViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+                    return
+                }
+                strongSelf.user = try result.unwrap()
+                strongSelf.name.value = strongSelf.user!.name
+                strongSelf.email.value = strongSelf.user!.email
+                debugPrint("Got user with id: " + strongSelf.user!.id!)
+            } catch let error as NSError {
```
mvazquez-ms: I'm not that familiar with Swift error handling, but seems like we are swallowing the error here, and that's not a good practice.

Once we do this:
> Create a service for authentication and remove references to Api

Let's implement proper error handling.

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/LoginViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+    
+    var name = Observable<String?>("")
+    var email = Observable<String?>("")
+    
+    init() {
+        // TODO: We need to implement dependency injection
+        self.api = Api.init()
+    }
+    
+    func getRandomUser() {
```
mvazquez-ms: Minor.. but you have two options here

```
func getRandomUser() {
   //get random Id
   //retrieve user with random Id
}
```

or 

```
func getUser(userId: String) {
   //retrieve user with userId
}

getUser(String(arc4random_uniform(10) + 1));
```

The first one is limiting the re usability of the class, only suited to get random users, that's kind of lame 😛

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/LoginViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+                strongSelf.email.value = strongSelf.user!.email
+                debugPrint("Got user with id: " + strongSelf.user!.id!)
+            } catch let error as NSError {
+                debugPrint("Get user error: \(error.localizedDescription)")
+            }
+        }
+    }
+    
+    func updateUser() {
+        if self.user != nil {
```
mvazquez-ms: [Golden Path](https://github.com/MakingSense/code-style-guides/blob/master/Swift/Readme.md#golden-path) 

paddlin

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/LoginViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+            self.user!.name = self.name.value!; // Should we bind the model with the viewModel?
```
mvazquez-ms: > // Should we bind the model with the viewModel?

Yep, in some scenarios that's a good idea.

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/LoginViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+            self.user!.email = self.email.value!;
```
mvazquez-ms: You already 'unwrapped' user value in the first `if`, you know it's not null, so you can avoid the multiple usages of `!` operator if you follow the Golden Path described on our style guide.

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/LoginViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+            api.createUser(user: self.user!) { [weak self] result in
+                do {
+                    guard let strongSelf = self else {
+                        return
+                    }
+                    strongSelf.user = try result.unwrap()
```
mvazquez-ms: WET code..

(Violations of DRY are typically referred to as WET solutions, which is commonly taken to stand for either "write everything twice", "we enjoy typing" or "waste everyone's time" - [source](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself))

You asked for a thorough review 😄 

This block is almost identical to the one above.

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/LoginViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+                    debugPrint("Updated user with id: " + strongSelf.user!.id!)
+                } catch let error as NSError {
+                    debugPrint("Update user error: \(error.localizedDescription)")
+                }
+            }
+        }
+    }
+    
+    func createUser() {
+        let user = User(name: "Mike Insense", email: "minsense@makingsense.com")
```
mvazquez-ms, line 68: > User(name: "Mike Insense"

Best user name ever!

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/LoginViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+        api.createUser(user: user) { [weak self] result in
+            do {
+                guard let strongSelf = self else {
+                    return
+                }
+                strongSelf.user = try result.unwrap()
```
mvazquez-ms: "we enjoy typing" code

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Models/Address.swift
```diff
@@ -0,0 +1,50 @@
...
+//  Created by Brian Sztamfater on 5/4/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+import SwiftyJSON
+
+struct Address {
+    
+    let street: String
```
mmaldini33: I do not understand why street is marked as let.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Api.swift
```diff
@@ -0,0 +1,64 @@
...
+//  Api.swift
+//  SwiftSeedProject
+//
+//  Created by Brian Sztamfater on 4/4/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+import Alamofire
+import SwiftyJSON
```
mmaldini33, line 11: cool!

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Models/Address.swift
```diff
@@ -0,0 +1,50 @@
...
+    init(json: JSON) {
+        self.street = json["street"].stringValue
+        self.suite = json["suite"].stringValue
+        self.city = json["city"].stringValue
+        self.zipcode = json["zipcode"].stringValue
+        self.location = Location(json: json["geo"])
+    }
+    
+    func toParameters() -> [String : AnyObject] {
+        var parameters = ["street" : street, "city" : city, "zipcode" : zipcode] as [String : Any]
```
mvazquez-ms: I think you have a couple of avoidable casts.

```
var parameters = ["street" : street, "city" : city, "zipcode" : zipcode] as [String : Any]
...
return parameters as [String : AnyObject]
```
Can be written as:

```
var parameters : [String : AnyObject] = ["street" : street, "city" : city, "zipcode" : zipcode]
...
return parameters
```

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Api.swift
```diff
@@ -0,0 +1,64 @@
...
+                case .failure(let error):
+                    completion(ApiResult{ throw error })
+            }
+        }
+    }
+    
+    func updateUser(user: User, completion: @escaping (ApiResult<User>) -> Void) {
+        let _ = manager.apiRequest(endpoint: .UpdateUser(userId: user.id!), parameters: user.toParameters()).apiResponse { response in
+            switch response.result {
+                case .success(let json):
```
mmaldini33: this code is identical to the getUser and createUser method. It will not be possible to centralize the code within the same method and depending on the result of each call, call the necessary method?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Api.swift
```diff
@@ -0,0 +1,64 @@
...
+                    let user = User(json: json)
+                    completion(ApiResult{ return user })
+                case .failure(let error):
+                    completion(ApiResult{ throw error })
```
mmaldini33: this code is identical to the getUser and createUser method. It will not be possible to centralize the code within the same method and depending on the result of each call, call the necessary method?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Endpoint.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+
+import Foundation
+import Alamofire
+
+enum Endpoint {
+    
+    case GetUser(userId: String)
+    case UpdateUser(userId: String)
+    case CreateUser()
```
mmaldini33, line 16: Great!

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Models/Location.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+import SwiftyJSON
+
+struct Location {
+    
+    var lat: String
+    var lng: String
```
mvazquez-ms: This is a matter of taste.. but most frameworks I've worked with used either Lat/Long or Latitude/Longitude.

'lng' doesn't feel right to me.. I prefer the full names, but in this case, at least add the 'o' between l and n, unless there's any reason not to.

mmaldini33 (commented):
Brian I left some comments on things that I think can be improved, but the reality is that the code looks pretty good

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Models/Location.swift
mvazquez-ms: And why are those strings? That's susceptible to localization errors when decimal separator is changed from comma to dot or vice versa. And it's hard to deal with comparisons/precision/thresholds (e.g. know if a certain location is inside a certain rectangle, it's a common scenario).

Lat and Long are usually defined as `double`.

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Models/User.swift
```diff
@@ -0,0 +1,53 @@
...
+//  SwiftSeedProject
+//
+//  Created by Brian Sztamfater on 4/4/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+import SwiftyJSON
+
+struct User {
```
mvazquez-ms: User should be a class, mostly because when you pass a user instance you expect a reference, not a copy of that user.

That's an important difference between struct and class.

mvazquez-ms (commented):
Just reviewed half the PR, need to go now..

Looking good! Left some comments.

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Models/Address.swift
```diff
@@ -0,0 +1,50 @@
...
+
+import Foundation
+import SwiftyJSON
+
+struct Address {
+    
+    let street: String
+    var suite: String?
+    var city: String
+    var zipcode: String
```
mvazquez-ms: Zip code should be an int.

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Models/User.swift
```diff
@@ -0,0 +1,53 @@
...
+    
+    var id: String?
```
mvazquez-ms: Is it OK to create a user without an `id`? According to this model, it is. According to usual business rules, it's not.

Model classes should always reflect business rules. There's a common scenario where a user with an empty id is required when creating a new user, and that's probably what your doing here, which is a code smell.

There are a couple of side-effects to that:

- Creating a user requires 'certain fields', usually name, email and just a couple more. 99% of the time, your business model will have more fields that those required to create a user. As a consumer of the `createUser` method receiving a business model `User` object, there's no way to know which fields are actually required on the create user operation, and which could be ignored.

- The reason why you are defining id as optional in a business model is a consequence of coupling business layer with third-party/external (from the PoV of this app) services. That coupling is not _that_ bad in the current code because you extracted the mapping into a separate method, protecting the code from contract changes on the API, but you can definitely see the influence of the API in your business model (in the Id field mostly).

- Consistency, for the same problem you want the same solution. Let's say we have an endpoint to edit a user (`EditUser`), and you need to provide only id for that endpoint (as required) and then any field to want to edit. So you send a `User` object as well, right? Just like `CreateUser`. Well, that means you'll have to change the already fragile constructor to something even more meaningless business-wise

`init(id: String? = nil, name: String, email: String, description: String? = nil, address: Address? = nil)`

`email `and `name `no longer can be mandatory. So this statement mentioned above:
> there's no way to know which fields are actually required on the create user operation, and which could be ignored

will hold true for almost every endpoint if you couple them with your business model. If you don't, then you have inconsistency, different solutions for the same problem. And if you want to consume REST API and provide parameters properly you should either check API's documentation (if any) or test until you find out (add and remove fields and check error code), beautiful!

- Using domain models as DTOs to consume APIs can easily lead to an[ anemic domain model](https://martinfowler.com/bliki/AnemicDomainModel.html). Having 'domain' objects with the sole purpose of transferring data.

How do you solve all these shit? By defining the data contract with the API in your code, at infrastructure level. There are multiple ways to achieve that, but I won't tell you any. 😄

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Models/User.swift
```diff
@@ -0,0 +1,53 @@
...
+    init(json: JSON) {
+        self.id = json["id"].stringValue
+        self.name = json["name"].stringValue
+        self.email = json["email"].stringValue
+        self.description = json["description"].string
+        self.address = Address(json: json["address"])
+    }
+    
+    func toParameters() -> [String : AnyObject] {
+        var parameters = ["name" : name, "email" : email] as [String : Any]
```
mvazquez-ms: You can avoid the two casts in this method as mentioned on a comment above.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Api.swift
mvazquez-ms: +1

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Api.swift
mvazquez-ms: +1

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Api.swift
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+    }
+    
+    func updateUser(user: User, completion: @escaping (ApiResult<User>) -> Void) {
+        let _ = manager.apiRequest(endpoint: .UpdateUser(userId: user.id!), parameters: user.toParameters()).apiResponse { response in
```
mvazquez-ms: Notice this subtlety:

`let _ = manager.apiRequest(endpoint: .UpdateUser(userId: user.id!)`,

You are forcing an unwrap (`!`) of user's id all over the code, when that should not be necessary at all. That's a code smell.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/ApiProtocol.swift
```diff
@@ -0,0 +1,89 @@
...
+    func apiRequest(endpoint: Endpoint, parameters: [String : AnyObject]?) -> ApiRequestProtocol {
+        return apiRequest(endpoint: endpoint, parameters: parameters, headers: nil)
+    }
+}
+
+protocol ApiRequestProtocol {
+    func apiResponse(completionHandler: @escaping (DataResponse<JSON>) -> Void) -> Self
+}
+
+func += <K, V> ( left: inout [K : V], right: [K : V]) {
```
mvazquez-ms: +1 for the operator overload
-1 for placing a completely generic and (I'm not 100% certain, but I think it is because it's not scoped by a class/protocol/struct/enum) global function in `ApiProtocol` file

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/ApiProtocol.swift
```diff
@@ -0,0 +1,89 @@
+//
```
mvazquez-ms: There are 2 protocols and 3 extensions in this file (as well as a global function). Do you think we can split them in at least two separate files (one for each protocol and its extensions)? And probably a third one for that function.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/ApiProtocol.swift
```diff
@@ -0,0 +1,89 @@
...
+    static func apiResponseSerializer() -> DataResponseSerializer<JSON> {
+        return DataResponseSerializer<JSON> { request, response, data, error in
+            if let error = error {
+                return .failure(error)
+            }
+            
+            guard let validData = data else {
+                let domain = Bundle.main.infoDictionary?[kCFBundleNameKey as String] as? String
+                let reason = "Data could not be serialized. Input data was nil."
+                return .failure(NSError(domain: domain!, code: 1001, userInfo: [NSLocalizedDescriptionKey : reason]))
```
mvazquez-ms: asdasdasdsadsa

On the first line the code is like "I don't know if we have a value here, so let's null-propagate everything", and the third line is like "Let's force-unwrap this shit, I'm mostly certain it's not null". 

The null-propagation make sense if then you add a guard/if checking nullity. Now it just delaying the inevitable.

For this code to make sense you should ether:

- Change the `as?` to `as!` (and it blow up in case something is null, that's what will happen with the current code anyway, but at least the operators will be consistent. Maybe Bundle.main.saraza is never null so it's OK to do `as!`)

- Add an if/guard check complementary to `as?` operator and then act accordingly if domain is null.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/ApiProtocol.swift
```diff
@@ -0,0 +1,89 @@
...
+            return sanitizeError(json: json)
+        }
+    }
+    
+    static func sanitizeError(json: JSON) -> Result<JSON> {
+        if !json["error"].exists() {
+            return .success(json)
+        }
+        
+        let code = json["error"]["code"].intValue
```
mvazquez-ms: From the perspective of a reusable class, this code is trouble.

You're assuming error messages from the API will have this specific format.  Even the same application can consume two different APIs with different error format.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Endpoint.swift
```diff
@@ -0,0 +1,53 @@
...
+            case .CreateUser():
+                return baseUrl.appendingPathComponent("users")!
+        }
+    }
+}
+
+private extension NSURL {
+    static func getBaseUrl() -> NSURL {
+        guard let info = Bundle.main.infoDictionary,
+            let urlString = info["Base URL"] as? String,
```
mvazquez-ms, line 46: Similar here, this implementation detail is hidden and could be annoying.

- Hidden because I will find out I have to define the base Url on the plist once I run a request and get this error. Although a developer can detect earlier that he have not provided a base Url anywhere and will have to look through the code and find this method (or give up and find the error later). This can be easily avoided by adding a explicit dependency somewhere so the first thing you have to do before instantiating whatever class is the initial point to this REST client design is to provide a base url. Then is up to the consumer if he want to retrieve that from the plist or anywhere else.

- There's no way to consume two different APIs within the same app with the current approach.

mvazquez-ms (changes requested):
Second part of the review.

I was multitasking (watching TV), so I might have misinterpreted your intentions/code.

Overall looks really good! Most of my comments are to generate debate and dig deeper into design concerns to be tackled for maintainable/reusable code.

I only expect the repeated code to be abstracted, and then the PR can be merged.

Everything else can be discussed separately and evaluate if it's worth it to change or not.

PR #16, opened 2017-04-05
MLP-154 - Integrate DI

lpelizzaMS:
# Background

We need to implement DI on our project. I have created two possible solutions one of them with Typhoon and the another with Dip. 
When I implemented the first one I found a limitation related to this -> "Every class you want to inject has to be a subclass of NSObject in some way (either by subclassing or adding @objc modifier)."
So, I chose the second option which doesn't have that limitation. 

# Changes done

I've implemented Dip which is a framework that solves the dependency injection paradigm. Also, I've included an example which uses the view controller and view model that already existed.

# Approvers required
@mvazquez-ms @mmaldini33  @jsmuguruza  @bsztamfater-ms

### DIFF
--- a/IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
+++ b/IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
@@ -7,13 +7,16 @@
 //
 
 import UIKit
+import Dip
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
-
+    private var container: DependencyContainer!
+    
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
+        container = DependencyContainer.configure()
         BuddyBuildSDK.setup()
         
         // Override point for customization after application launch.
--- a/IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/LoginViewController.swift
+++ b/IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/LoginViewController.swift
@@ -8,6 +8,7 @@
 
 import UIKit
 import Bond
+import Dip_UI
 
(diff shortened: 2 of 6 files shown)


mmaldini33 (approved):

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/LoginViewController.swift
```diff
@@ -32,3 +33,5 @@ class LoginViewController: UIViewController {
     }
 
 }
+
+extension LoginViewController: StoryboardInstantiatable { }
```
mvazquez-ms, line 37: What's the purpose of this? Is it required by Dip?

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
```diff
@@ -7,12 +7,14 @@
...
 
 import UIKit
+import Dip
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
-
+    private let container = DependencyContainer.configure()
```
mvazquez-ms: I think this line should be inside `didFinishLaunchingWithOptions `. The composition root could grow considerably and should only be built if the app was initialized successfully. With the current code, you have no control over that.

mvazquez-ms (approved):
Awesome simple PR!

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/LoginViewController.swift
lpelizzaMS: This is required by Dip, it's needed for the instantiation of VC.

IOS/SwiftSeedProject/SwiftSeedProject/AppDelegate.swift
lpelizzaMS: roger!

briansztamfater (approved):

PR #17, opened 2017-04-13
[MLP-160] Integrate newsapi.org API

briansztamfater:
### Background

We need to integrate NewsApi.org API

### Changes done

Created all the endpoints and models, and leaved a short implementation on LoginViewModel (should be completely refactored and decoupled on its respective components and services when @lpelizzaMS rebases his changes)

### Approvers required

@lpelizzaMS @mvazquez-ms @mmaldini33 

### Optional approvers

@jsmuguruza


IOS/SwiftSeedProject/SwiftSeedProject/Domain/Article.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Created by Brian Sztamfater on 13/4/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+import SwiftyJSON
+
+class Article {
+    
+    var author: String?
```
mvazquez-ms: I'm not sure.. but you should be able to define all these fields as `let` right? Since you're only setting them inside the `init` method.

Also, can you define them as mandatory instead of optional?

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Source.swift
```diff
@@ -0,0 +1,29 @@
...
+import Foundation
+import SwiftyJSON
+
+class Source {
+    
+    var id: String!
+    var name: String?
+    var description: String?
+    var url: URL?
+    var language: String?
```
mvazquez-ms: Can you define this as `var language : Languate`  ? (since you already defined the enum)

Also, check if these fields can be defined as `let` too. Such values shouldn't change once instantiated.

IOS/SwiftSeedProject/SwiftSeedProject/Info.plist
```diff
@@ -4,6 +4,8 @@
 <dict>
 	<key>Base URL</key>
 	<string>$(BASE_URL)</string>
+	<key>API Key</key>
```
mvazquez-ms: Can you add something more to this name? Something like `newsapi API key` or ``news API key`

(We can easily have multiple api keys)

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Extensions/DataRequest+Extensions.swift
```diff
@@ -30,13 +30,12 @@ extension DataRequest: ApiRequestProtocol {
     }
     
     static func sanitizeError(json: JSON) -> Result<JSON> {
-        if !json["error"].exists() {
+        if json["status"].stringValue == "ok" {
```
mvazquez-ms, line 33: If we want to check for Http Status code, where can we obtain it?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Extensions/DataRequest+Extensions.swift
```diff
@@ -30,13 +30,12 @@ extension DataRequest: ApiRequestProtocol {
...
             return .success(json)
         }
 
-        let code = json["error"]["code"].intValue
-        let message = json["error"]["message"].stringValue
-        let error = NSError(domain: bundleNameKey(), code: code, userInfo: [NSLocalizedDescriptionKey : message])
+        let message = json["message"].stringValue
+        let error = NSError(domain: bundleNameKey(), code: -9999, userInfo: [NSLocalizedDescriptionKey : message])
```
mvazquez-ms: Why the -9999? 😄

mvazquez-ms (commented):
Just some minor comments..

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Extensions/DataRequest+Extensions.swift
briansztamfater: The closure that calls this function has a reference to the response object, which has a statusCode property. On this method we just parse the error if exists

IOS/SwiftSeedProject/SwiftSeedProject/Info.plist
briansztamfater: Yep!

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Source.swift
briansztamfater: Sure, my fault

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Extensions/DataRequest+Extensions.swift
briansztamfater: This API doesn't return an error code, so I just hardcoded one. I should add an enum for this I think

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Api.swift
```diff
@@ -13,13 +13,22 @@ import SwiftyJSON
...
-    private let processUser: (Result<JSON>, (ApiResult<UserM>) -> Void) -> Void = { result, completion in
+    private let processSources: (Result<JSON>, (ApiResult<Array<Source>>) -> Void) -> Void = { result, completion in
         switch result {
-            case .success(let json):
-                let user = UserM(json: json)
-                completion(ApiResult{ return user })
-            case .failure(let error):
-                completion(ApiResult{ throw error })
+        case .success(let json):
+            let sources = json["sources"].arrayValue.map { Source(json: $0) }
```
briansztamfater: Love this

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Language.swift
```diff
@@ -0,0 +1,15 @@
...
+//  Language.swift
+//  SwiftSeedProject
+//
+//  Created by Brian Sztamfater on 13/4/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+
+enum Language: String {
```
lpelizzaMS, line 11: Are you sure that these are all languages?

lpelizzaMS (commented):
👍

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Language.swift
briansztamfater: From the API documentation (https://newsapi.org/#apiSources): 

> Possible options: en, de, fr.

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/BaseService.swift
```diff
@@ -7,25 +7,34 @@
...
     public func getAll() -> [T] {
         return persistence.getAll(entityName: T.EntityName)
     }
     
-    public func getEntityBy(identifier: Int) -> T {
+    public func getEntityBy(identifier: String) -> T? {
         return persistence.getBy(entityIdentifier: identifier, entityName: T.EntityName)
     }
     
+    public func createEntity(json: JSON) {
```
mvazquez-ms: This is a bit weird..

So, we have

App
'--------'
Domain
'--------'
Infra

`json` belongs to infra, but this class is exposing a method to be used by `App` layer (because public methods are consumed by the upper layer, otherwise they should be protected/private/internal-C#-style). Anyone consuming this method outside a service will be breaking the layered architecture.

On the other hand, it seems you are using it only within Domain layer, and more precisely, inside a child class,, so this method should be protected, and that way we prevent weird scenarios were a ViewModel ends up handling `json`.

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/SourceService.swift
```diff
@@ -7,10 +7,31 @@
...
+            do {
+                guard let strongSelf = self else {
+                    return
+                }
+                let json = try result.unwrap()
+                let sources = json["sources"].arrayValue
+                sources.forEach { json in
+                    let id = json["id"].stringValue
+                    if let source = strongSelf.getEntityBy(identifier: id) as Source? {
+                        source.updateWithJSON(json: json)
```
mvazquez-ms: Mmm.. about these lines:

`source.updateWithJSON(json: json)`
`strongSelf.createEntity(json: json)`

The call to `updateWithJSON` will update the dabatase right-away? (or only in-memory?) If that's not the case, then maybe we should delegate the update into another class (the same that defines `createEntity`).

mvazquez-ms: Oh, nevermind, just saw the `strongSelf.save()` call at the end of the loop.

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/SourceListViewModel.swift
```diff
@@ -0,0 +1,47 @@
...
+class SourceListViewModel {
+    
+    private let sourceService: SourceService!
+    
+    let sources = MutableObservableArray<SourceViewModel>()
+    
+    init(sourceService: SourceService) {
+        self.sourceService = sourceService
+        self.setupNotifications()
+        self.getSources()
```
mvazquez-ms: It's usually a bad practice to do stuff other than configuration/setup inside the constructor. Calling `getSources()` and `requestSourcesFromServer()` on the constructor can lead to unwanted behavior.

Constructing an object should fail only on exceptional cases (like trying to initialize it with wrong parameters). Here you have a lot of reasons for the constructor to fail (e.g. CoreData, parsing, network, API), the constructor cannot handle any of those error, and you can't expect consumers to handle such errors when instantiating this ViewModel.

Also, you are hiding behavior. Consumer won't expect this class to do requests to an API upon instantiation.

You could do that more evident by adding an extra parameter on the constructor like `preload` or something like that, but I think it's better to provide methods so consumers can perform those actions on demand.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/DateUtils.swift
```diff
@@ -0,0 +1,23 @@
...
+
+import Foundation
+
+class DateUtils {
+    
+    struct Format {
+        static let ISO8601 = "yyyy-MM-dd'T'HH:mm:ssZ"
+    }
+    
+    static func parse(dateString: String, format: String) -> Date {
```
mvazquez-ms: What do you think about using ISO8601 as default value for `format ` parameter?

mvazquez-ms (commented):
Just some minor comments, overall looks great!

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/DateUtils.swift
briansztamfater: Great idea. Done!

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/BaseService.swift
```diff
@@ -7,25 +7,34 @@
...
+        persistence.add(json: json, entityName: T.EntityName)
+    }
+    
+    public func save() {
+        persistence.save()
```
lpelizzaMS: I think that the service should not has a save method. What do you think?

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/BaseService.swift
lpelizzaMS: I think that the Service layer should not be able to save from json. Maybe the api should has some way to save on core data.

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/BaseService.swift
```diff
@@ -7,25 +7,34 @@
...
     
-    private let persistence: Persistence
-    private let restClient: AnyObject? = nil
-    
-    //TODO: we should inclued the API access also
-    init(persistence: Persistence) {
+    private let persistence: Persistence!
+    internal let api: Api!
+
+    init(persistence: Persistence, api: Api) {
```
lpelizzaMS: I think that API is not a good name. I am awful to create names, but maybe @mvazquez-ms  can help us with this.

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/BaseService.swift
```diff
@@ -7,25 +7,34 @@
 //
 
 import Foundation
+import SwiftyJSON
```
lpelizzaMS: I think that Service should not has this kind of imports

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/SourceService.swift
lpelizzaMS: I agree with marce comment, maybe we should create another class to delegate some logic

lpelizzaMS (changes requested):
it's great!
I think that we should do some little changes before merge. Maybe Marcelo can help us with those changes

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/BaseService.swift
briansztamfater: I don't think the network layer should persist objects, because it would be mixing responsibilities and also creating a new dependency on persistance on the Network layer. The service is already delegating the save to the Persistence. What do you think? Maybe I'm not getting something

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/BaseService.swift
briansztamfater: Another option would be to make the persistance object protected and call save from the persistance directly. But I wanted to be consistent with the other functions.

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/BaseService.swift
briansztamfater: Yep I don't like it either, I bet @mvazquez-ms will come with a good alternative

mvazquez-ms: Yep, `Api` is way too generic. Maybe `RestApiClient`? or just `RestClient`? for the class name.

And that particular instance should be called something like `newsApiRestClient`

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/BaseService.swift
mvazquez-ms: > I think that the service should not has a save method. What do you think?

Exposed as a public method? definitely not. Good catch!

> Another option would be to make the persistence object protected and call save from the persistence directly

Yes, I think that's OK. It might look like an inconsistency, but once this services grow and handle all the domain-related stuff, it will make much more sense. Now it looks like a simple persistence proxy

mmaldini33 (commented):
I Will wait the changes to review again the code.

IOS/SwiftSeedProject/SwiftSeedProject/Application/Service/BaseService.swift
mvazquez-ms: Yep, ideally, not.

In this case, we have 

`internal func createEntity(json: JSON)`

A proper interface for a domain service should be:

`internal func createEntity(entity: T)`

And the same happens with persistence class.. instead of:

`func add(json: JSON, entityName: String)`

we should have

`func add( entity: T, entityName: String)`

JSON usage should be contained within `RectClient` related classes. Once a response is received, it's parsed into an entity or DTO, and then handled by persistence or domain services.

I'm not sure how difficult that could be with CoreData, do you think we can change it?

IOS/SwiftSeedProject/SwiftSeedProject/Application/SourceViewModel.swift
```diff
@@ -0,0 +1,35 @@
...
+    let identifier: Observable<String>
+    let name: Observable<String>
+    let url: Observable<URL>
+    let detail: Observable<String>
+    let language: Observable<Language>
+    let country: Observable<Country>
+    let category: Observable<SourceCategory>
+    
+    init(source: Source) {
+        self.source = source
```
lpelizzaMS: do we need to have the property source? I am asking because i don't know how works Bond.

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Article+CoreDataProperties.swift
```diff
@@ -19,8 +19,8 @@ extension Article {
     @NSManaged public var author: String?
     @NSManaged public var title: String?
     @NSManaged public var detail: String?
-    @NSManaged public var url: String?
-    @NSManaged public var urlToImage: String?
+    @NSManaged public var url: URL?
```
lpelizzaMS, line 22: did you change the dataModel?

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Article.swift
```diff
@@ -0,0 +1,29 @@
...
+//  SwiftSeedProject
+//
+//  Created by Brian Sztamfater on 13/4/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+import SwiftyJSON
+
+class ArticleM {
```
lpelizzaMS: @bsztamfater-ms  & @mvazquez-ms  We will work with 2 models for each entity?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
```diff
@@ -0,0 +1,78 @@
...
+
+import Foundation
+import Alamofire
+import SwiftyJSON
+
+final class RestClient {
+    // MARK: - Private Properties
+    private let manager: ApiManagerProtocol
+    private let persistence: Persistence
+    private let processSources: (Persistence, Result<JSON>, (ApiResult<Array<Source>>) -> Void) -> Void = { persistence, result, completion in
```
lpelizzaMS, line 17: Could we try to create a generic method, couldn't we? What do you think?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
```diff
@@ -0,0 +1,78 @@
...
+                }
+                persistence.save()
+                completion(ApiResult{ return articles })
+            case .failure(let error):
+                completion(ApiResult{ throw error })
+        }
+    }
+    
+    // MARK: - Designated Initializer
+    init(manager: ApiManagerProtocol = Configuration.manager, persistence: Persistence) {
```
lpelizzaMS: why are you using Configuration.manager on init?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
```diff
@@ -0,0 +1,78 @@
...
+        self.manager = manager
+        self.persistence = persistence
+    }
+    
+    // MARK: - Public Methods
+    func getSources(category: SourceCategory? = nil, language: Language? = nil, country: Country? = nil, completion: @escaping (ApiResult<Array<Source>>) -> Void) {
```
lpelizzaMS, line 54: the same question then before, this looks like similar code all the time. but what do you think? maybe I am missing something

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
```diff
@@ -0,0 +1,78 @@
...
+            guard let strongSelf = self else {
+                return
+            }
+            strongSelf.processArticles(strongSelf.persistence, response.result, completion)
+        }
+    }
+    
+}
+
+struct Configuration {
```
lpelizzaMS: why is this struct on the same file than the class RestClient?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/PersistenceData.swift
```diff
@@ -20,54 +20,58 @@ class PersistenceData: Persistence {
     }
     
     public func getAll<T: PersistenceObject>(entityName: String) -> [T] {
-        let objectContext = coreDataManager.privateChildManagedObjectContext()
+        let objectContext = coreDataManager.mainManagedObjectContext
```
lpelizzaMS, line 23: why did you change the privateChildManagedObjectContext by MainManagedObjectContext?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/PersistenceData.swift
```diff
@@ -20,54 +20,58 @@ class PersistenceData: Persistence {
...
         }
         return []
     }
     
-    public func getBy<T: PersistenceObject>(entityIdentifier: Int, entityName: String) -> T {
-        let objectContext = coreDataManager.privateChildManagedObjectContext()
-        let fetch = NSFetchRequest<NSFetchRequestResult>(entityName: entityName)
-        fetch.predicate = NSPredicate(format: "identifier == %@", entityIdentifier)
+    public func getBy<T: PersistenceObject>(entityIdentifier: String, entityName: String) -> T? {
+        let objectContext = coreDataManager.mainManagedObjectContext
```
lpelizzaMS, line 36: same Q than before, why did you change the child MOC by the main?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/PersistenceData.swift
```diff
@@ -20,54 +20,58 @@ class PersistenceData: Persistence {
...
+            return fetchedObjects.count > 0 ? fetchedObjects.item(at: 0) as? T : nil
         } catch let error as NSError {
             fatalError("Failed to fetch \(entityName): \(error)")
-            print("Could not getBy. \(error), \(error.userInfo)")
         }
     }
     
-    public func add(attributes: [String : AnyObject], entityName: String) {
-        let objectContext = coreDataManager.privateChildManagedObjectContext()
+    public func add(attributes: [String : AnyObject]? = nil, entityName: String) -> PersistenceObject {
```
lpelizzaMS: why didn't you use:
func add<T: PersistenceObject>(attributes: [String : AnyObject]? = nil, entityName: String) -> T
instead of :
func add(attributes: [String : AnyObject]? = nil, entityName: String) -> PersistenceObject

lpelizzaMS (changes requested):
It looks great! I wrote some minor comments

IOS/SwiftSeedProject/SwiftSeedProject/Application/SourceViewModel.swift
briansztamfater: Not necessary at all, we can remove it. The binded properties are the Observable ones.

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Article.swift
briansztamfater: We can remove this class as we are now using Core Data entities. @mvazquez-ms is it ok to delete it?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
briansztamfater: Yep, but to move on by now I would add a TODO comment, and then refactor this (there are a few thing that we need to improve on RestClient)

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
briansztamfater: As I pointed above, we need a bit of refactor of this class, but I prefer to move on and refactor it when we have more progress on the seed.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/PersistenceData.swift
briansztamfater: Context on new objects was nil after returning the object, so the update was failing

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/PersistenceData.swift
briansztamfater: Idem

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/PersistenceData.swift
briansztamfater: Actually I did it that way firstly, but then changed to this one. No specific reason lol. Which advantages and disadvantages we have with the first option over the second one?

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
briansztamfater: Good point, I forgot to move it to a new file when I first implemented the RestClient. Will do.

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Article+CoreDataProperties.swift
briansztamfater: Yes, just to use URLs instead of Strings. I marked those attributes as Transformables

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
briansztamfater: We need it to make requests using the Endpoint enum

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/PersistenceData.swift
lpelizzaMS: I think that when you use the first one  you don't need to do a casting

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/PersistenceData.swift
lpelizzaMS: we need to review this because on this new way you are using always the main thread which means that in some cases you are blocking this.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
lpelizzaMS: I totally agree with you that those changes need to be on other PR. But I think that those changes need to create next to this one. we should complete the rest client asap. what do   @mvazquez-ms think?

IOS/SwiftSeedProject/SwiftSeedProject/Application/SourceViewModel.swift
lpelizzaMS: great!

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
mmaldini33: I consider that finish the rest client asap is a good idea!

IOS/SwiftSeedProject/SwiftSeedProject/Domain/Article.swift
mvazquez-ms: Sure, go ahead, I'm always in favor of removing stuff!

mmaldini33: lol

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/RestClient.swift
mvazquez-ms: Yep, let's wrap these changes and create a separate PR for the refactors discussed here.

PR #18, opened 2017-04-13
[MLO-159]Create models & services

lpelizzaMS:
# Background

We need to create the models to save locally the information from NewsApi.org. Also, we should have the service which will be used by VM.

# Changes done

Created all models and services to work with the NewsApi.org's information.

# Approvers required

@bsztamfater-ms  @mvazquez-ms  @mmaldini33  @jsmuguruza


mvazquez-ms (approved):
Awesome PR @lpelizzaMS !

briansztamfater (approved):

mmaldini33 (commented):
Excellent!

PR #19, opened 2017-04-19
[MLP-162]Create news feed screen

lpelizzaMS:
# Background

We need to create a new screen to show the news

# Changes done

Created all the components that we need to show the news. They are:
- A screen on the storyboard.
- A view controller
- A collection view
- A custom cell for collection view
- Some protocols
Also, I've added a new pod which is SDWebImage. This pod is a great option to load images.

# Approvers required

@bsztamfater-ms  @mvazquez-ms @mmaldini33


IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/SSNewsCollectionCellVM.swift
```diff
@@ -0,0 +1,29 @@
...
+//  SSNewsCollectionCellVM.swift
+//  SwiftSeedProject
+//
+//  Created by Lucas Pelizza on 4/18/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+
+final class SSNewsCollectionCellVM {
```
mvazquez-ms: Adding prefixes to classes was common on Obj-c (because there were no namespaces), but I think we should avoid it on Swift. It's part of our guidelines [here](https://github.com/MakingSense/code-style-guides/blob/master/Swift/Readme.md#class-prefixes).

The MS prefix (e.g. `MSButton`) could be an exception because we want to make it very obvious it's part of our 'framework/reusable classes'.

So, unless there's a compelling reason, I prefer we avoid prefixes on Swift.

mvazquez-ms: Also, `NewsCollectionCellVM` doesn't feel like an appropriate name for a `ViewModel`, it's tied to UI.

For what I can see here, this ViewModel represents a piece of news, and 'news' is a tricky word (an irregular plural), so maybe we can start using the name 'Article', and this would be an `ArticleViewModel`or `NewsArticleViewModel`

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/SSNewsCollectionCellVM.swift
```diff
@@ -0,0 +1,29 @@
...
+    
+    let imageUrl: String
+    let date: Date
+    let description: String
+    let title: String
+    let author: String
+    let sourceURL: String
```
mvazquez-ms: I saw @bsztamfater-ms defined URLs as `URL` type instead of `string`, check with him so we take a consistent approach, and user one or the other.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/SSNewsCollectionViewCell.swift
```diff
@@ -0,0 +1,45 @@
...
+    @IBOutlet weak var lblDate: UILabel!
+    @IBOutlet weak var lblDescription: UILabel!
+    @IBOutlet weak var lblTitle: UILabel!
+    @IBOutlet weak var lblAuthor: UILabel!
+    
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        self.contentView.frame = CGRect(x: self.contentView.frame.origin.x, y: self.contentView.frame.origin.y, width:UIScreen.main.bounds.width , height: CGFloat(SSNewsCollectionViewCell.Height))
+    }
+
```
mvazquez-ms: Additional blank line.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/SSNewsCollectionViewCell.swift
```diff
@@ -0,0 +1,45 @@
...
+    
+    @IBOutlet weak var imgDetail: UIImageView!
+    @IBOutlet weak var lblDate: UILabel!
+    @IBOutlet weak var lblDescription: UILabel!
+    @IBOutlet weak var lblTitle: UILabel!
+    @IBOutlet weak var lblAuthor: UILabel!
+    
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        self.contentView.frame = CGRect(x: self.contentView.frame.origin.x, y: self.contentView.frame.origin.y, width:UIScreen.main.bounds.width , height: CGFloat(SSNewsCollectionViewCell.Height))
```
mvazquez-ms: Is this required? Shouldn't be solved by the `getHeightForRow` (or something like that) method?

And overriding the width with `UIScreen.main.bounds.width` shouldn't be required.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/SSNewsCollectionViewCell.swift
```diff
@@ -0,0 +1,45 @@
...
+        lblDate.text = "\(viewModel.date)"
+        lblDescription.text = viewModel.description
+        lblTitle.text = viewModel.title
+        lblAuthor.text = viewModel.author
+    }
+    
+    override func layoutSubviews()
+    {
+        imgDetail.layer.borderWidth = 1
+        imgDetail.layer.borderColor = UIColor.clear.cgColor
```
mvazquez-ms: Is this a transparent border with width 1? Is it OK? Sounds weird 😄 

(If you want a separator border to differentiate the image from the background, I think it should have a color, not transparent, but I'm not sure what was the intention here)

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/SSNewsCollectionView.swift
```diff
@@ -0,0 +1,55 @@
...
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import UIKit
+
+final class SSNewsCollectionView: UICollectionView, UICollectionViewDelegate, UICollectionViewDataSource {
+
+    var items: [SSNewsCollectionCellVM] = []
+    weak var viewModelDelegate: SSNewsCollectionViewModel?
+    var cellIdentifier: String?
```
mvazquez-ms: If this class will use only `SSNewsCollectionViewCell` cells, does it needs to store cellIdentifier here? Shouldn't it be using the const defined on `SSNewsCollectionViewCell`?

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/SSNewsCollectionView.swift
```diff
@@ -0,0 +1,55 @@
...
+        self.items = []
+        self.viewModelDelegate = viewModelDelegate
+        self.dataSource = self
+    }
+    
+    func setItems(_ items: [SSNewsCollectionCellVM]) {
+        self.items = items
+        self.reloadData()
+    }
+
```
mvazquez-ms: Additional white line.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/SSNewsFeedViewController.swift
```diff
@@ -0,0 +1,43 @@
...
+    
+    @IBOutlet weak var cvNews: SSNewsCollectionView!
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        //TODO: here we should init the VM
+        self.cvNews.setup(viewModelDelegate: self.viewModel! as! SSNewsCollectionViewModel, cellIdentifier: SSNewsCollectionViewCell.identifier)
+    }
+
+    override func didReceiveMemoryWarning() {
```
mvazquez-ms: You can safely remove this method.

mvazquez-ms (commented):
Awesome! You're doing a great job! Just left some comments, but the code is mergeable if required.

briansztamfater (changes requested):
Good job @lpelizzaMS! The only thing I would change is to remove the Pods folder from the commit (isn't it added to the .gitignore file? That's weird)

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/SSNewsCollectionCellVM.swift
lpelizzaMS: Yes, I know that the singular is: a piece of news. But the name will be very long. I will use one of the options you've mentioned.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/SSNewsCollectionViewCell.swift
lpelizzaMS: Yup, I've forgotten assign the color.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/SSNewsCollectionView.swift
lpelizzaMS: Yes, I had started to create the class more generic but little by little I was changing. I will remove the "cellIdentifier"

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsArticleViewModel.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+//  Created by Lucas Pelizza on 4/18/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+
+final class NewsArticleViewModel {
+    
+    let imageUrl: String
```
briansztamfater: We should use Bond's `Observable<T>` instead of just the type (i.e `Observable<String>`) so we can bind the properties on the View

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsCollectionViewCell.swift
```diff
@@ -0,0 +1,42 @@
...
+    @IBOutlet weak var lblDescription: UILabel!
+    @IBOutlet weak var lblTitle: UILabel!
+    @IBOutlet weak var lblAuthor: UILabel!
+    
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        self.contentView.frame = CGRect(x: self.contentView.frame.origin.x, y: self.contentView.frame.origin.y, width:UIScreen.main.bounds.width , height: CGFloat(NewsCollectionViewCell.Height))
+    }
+
+    func setup(viewModel: NewsArticleViewModel) {
```
briansztamfater: We should use data binding here

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/NewsCollectionView.swift
```diff
@@ -0,0 +1,52 @@
...
+//
+//  Created by Lucas Pelizza on 4/18/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import UIKit
+
+final class NewsCollectionView: UICollectionView, UICollectionViewDelegate, UICollectionViewDataSource {
+
+    var items: [NewsArticleViewModel] = []
```
briansztamfater: We should use Bond's `MutableObservableArray<NewsArticleViewModel>` so we can bind it

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/NewsFeedViewController.swift
```diff
@@ -0,0 +1,43 @@
...
+    
+    @IBOutlet weak var cvNews: NewsCollectionView!
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        //TODO: here we should init the VM
+        self.cvNews.setup(viewModelDelegate: self.viewModel! as! NewsViewModel)
+    }
+
+    override func didReceiveMemoryWarning() {
```
briansztamfater: This method should be removed if its implementation is empty

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/NewsFeedViewController.swift
```diff
@@ -0,0 +1,43 @@
...
+        super.didReceiveMemoryWarning()
+    }
+    
+    //MARK: - SSNewsFeedVMDelegate
+    func updateUI(news: [NewsArticleViewModel]) {
+        cvNews.setItems(news)
+    }
+    
+    func navigateTo(vc:UIViewController) {
```
briansztamfater: I think this function should be a protected function located in a BaseViewController so is available to every ViewController that extends it. What do you think?

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/NewsCollectionView.swift
```diff
@@ -0,0 +1,52 @@
...
+    weak var viewModelDelegate: NewsViewModel?
+    
+    func setup(viewModelDelegate: NewsViewModel)
+    {
```
briansztamfater: From the style guide:

> Method braces and other braces (if/else/switch/while etc.) always open on the same line as the statement but close on a new line.
> 
> **Preferred:**
> `if user.isHappy {`
>  `// Do something`
> `} else {`
>  `// Do something else`
> `}`
>
> **Not Preferred:** 
> `if user.isHappy`
> `{`
>  `// Do something`
> `}`
> `else {`
> `// Do something else`
> `}`

briansztamfater:
Leaved some comments, mostly minor and easy to fix

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/NewsCollectionView.swift
```diff
@@ -0,0 +1,52 @@
...
+    }
+    
+    func setItems(_ items: [NewsArticleViewModel]) {
+        self.items = items
+        self.reloadData()
+    }
+    
+    //MARK: - UICollectionViewDelegate
+    func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath)
+    {
```
mmaldini33: Method braces and other braces (if/else/switch/while etc.) always open on the same line as the statement but close on a new line.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/NewsCollectionView.swift
```diff
@@ -0,0 +1,52 @@
...
+        viewModelDelegate?.selectedRow(rowVM: (items[indexPath.row]))
+    }
+    
+    //MARK: - UICollectionView Data Source
+    func numberOfSections(in collectionView: UICollectionView) -> Int
+    {
```
mmaldini33: Method braces and other braces (if/else/switch/while etc.) always open on the same line as the statement but close on a new line.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/NewsCollectionView.swift
```diff
@@ -0,0 +1,52 @@
...
+        return 1
+    }
+    
+    func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int
+    {
```
mmaldini33: Method braces and other braces (if/else/switch/while etc.) always open on the same line as the statement but close on a new line.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/NewsCollectionView.swift
```diff
@@ -0,0 +1,52 @@
...
+        return items.count
+    }
+    
+    func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell
+    {
```
mmaldini33: Method braces and other braces (if/else/switch/while etc.) always open on the same line as the statement but close on a new line.

mmaldini33 (approved):
Minor comments

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsArticleViewModel.swift
mvazquez-ms: I thought the same when I first saw this ViewModel, but usually you only define observable properties for values you know will change through object's life-cycle.

In this particular case, none of this values will change after the object is instantiated, they will never trigger a change notification, so making these properties observable is not strictly required (unless the frameworks we are using requires that, but that would be odd. They should support a classic one-time binding).

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsCollectionViewCell.swift
mvazquez-ms: +1

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsCollectionViewCell.swift
```diff
@@ -0,0 +1,42 @@
...
+    
+    @IBOutlet weak var imgDetail: UIImageView!
+    @IBOutlet weak var lblDate: UILabel!
+    @IBOutlet weak var lblDescription: UILabel!
+    @IBOutlet weak var lblTitle: UILabel!
+    @IBOutlet weak var lblAuthor: UILabel!
+    
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        self.contentView.frame = CGRect(x: self.contentView.frame.origin.x, y: self.contentView.frame.origin.y, width:UIScreen.main.bounds.width , height: CGFloat(NewsCollectionViewCell.Height))
```
mvazquez-ms: Since after renaming the file, the comment is lost, I'll ask again. 

Do we need this line? The height should be resolved via `getHeightForRow`, and settings width to `UIScreen.main.bounds.width` shouldn't be required.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/NewsCollectionView.swift
mvazquez-ms: +1

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/NewsCollectionView.swift
```diff
@@ -0,0 +1,52 @@
...
+        let cell = collectionView.dequeueReusableCell(withReuseIdentifier: NewsCollectionViewCell.identifier, for: indexPath as IndexPath) as! NewsCollectionViewCell
+        cell.setup(viewModel:items[indexPath.row])
+        return cell as UICollectionViewCell
```
mvazquez-ms: This up-cast shouldn't be required.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsCollectionViewCell.swift
lpelizzaMS: Why do you want to use data binding here? This is the same case that the previous comment
@bsztamfater-ms & @mvazquez-ms

mvazquez-ms: > This is the same case that the previous comment @bsztamfater-ms & @mvazquez-ms

My comment was about defining properties as observable or not, that's related but no the same as binding.

To be consistent through out the app, I think we should always communicate views and data in the same way.  Here you will have one-time binding (evaluated only when the view is created, and they don't require change notifications).

We'll definitely use binding in many places, so to keep consistency, it's better to use it everywhere. And chances are most ViewModels will have notifiable properties.. a NewsArticle can have a `comments` or `likes` indicator that should be notifiable to be update in real time.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/NewsFeedCollectionViewController.swift
```diff
@@ -0,0 +1,38 @@
...
+import Bond
+import Dip_UI
+import Dip
+
+private let reuseIdentifier = "NewsCollectionViewCell"
+
+class NewsFeedCollectionViewController: UICollectionViewController {
+
+    var viewModel: NewsFeedViewModel!
+    var dataSource: MutableObservableArray<MutableObservableArray<NewsArticleViewModel>>!
```
briansztamfater: Why is it a MutableObservableArray of MutableObservableArrays? Couldn't you bind viewModel.newsArticles directly?

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsArticleViewModel.swift
```diff
@@ -0,0 +1,29 @@
...
+//  SwiftSeedProject
+//
+//  Created by Lucas Pelizza on 4/18/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+import Bond
+
+final class NewsArticleViewModel {
```
briansztamfater: I would name it ArticleViewModel as it is a ViewModel of Article. What do you think? Just a suggestion

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/SourceListViewController.swift
```diff
@@ -35,4 +35,4 @@ class SourceListViewController: UIViewController {
 
 }
 
-extension SourceListViewController: StoryboardInstantiatable { }
+//extension SourceListViewController: StoryboardInstantiatable { }
```
briansztamfater: Why is this line commented?

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/NewsFeedCollectionViewController.swift
lpelizzaMS: The Bond example was like that.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsArticleViewModel.swift
lpelizzaMS: great

IOS/SwiftSeedProject/SwiftSeedProject/Application/ViewModels/NewsFeedViewModel.swift
```diff
@@ -0,0 +1,47 @@
...
+import Bond
+
+class NewsFeedViewModel {
+    
+    private let libraryService: LibraryService!
+    let newsArticles = MutableObservableArray<ArticleViewModel>()
+    
+    init(libraryService: LibraryService) {
+        self.libraryService = libraryService
+        self.requestSourcesFromServer()
```
mvazquez-ms: Same comment as @bsztamfater-ms 's PR

>It's usually a bad practice to do stuff other than configuration/setup inside the constructor. Calling getSources() and requestSourcesFromServer() on the constructor can lead to unwanted behavior.

>Constructing an object should fail only on exceptional cases (like trying to initialize it with wrong parameters). Here you have a lot of reasons for the constructor to fail (e.g. CoreData, parsing, network, API), the constructor cannot handle any of those error, and you can't expect consumers to handle such errors when instantiating this ViewModel.

>Also, you are hiding behavior. Consumer won't expect this class to do requests to an API upon instantiation.

>You could do that more evident by adding an extra parameter on the constructor like preload or something like that, but I think it's better to provide methods so consumers can perform those actions on demand.

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/PersistenceData.swift
```diff
@@ -32,6 +32,19 @@ class PersistenceData: Persistence {
         return []
     }
     
+    public func getCustomBy<T: PersistenceObject>(attributeName: String, attributeValue: String) -> T? {
```
mvazquez-ms: WET code.. almost same code as 
```Swift
getBy<T: PersistenceObject>(entityIdentifier: String, entityName: String) -> T?
```

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsCollectionViewCell.swift
```diff
@@ -0,0 +1,44 @@
...
+    @IBOutlet weak var imgDetail: UIImageView!
+    @IBOutlet weak var lblDate: UILabel!
+    @IBOutlet weak var lblDescription: UILabel!
+    @IBOutlet weak var lblTitle: UILabel!
+    @IBOutlet weak var lblAuthor: UILabel!
+
+    
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        self.contentView.frame = CGRect(x: self.contentView.frame.origin.x, y: self.contentView.frame.origin.y, width:UIScreen.main.bounds.width , height: CGFloat(NewsCollectionViewCell.Height))
```
mvazquez-ms: Same question again 😄 

> Since after renaming the file, the comment is lost, I'll ask again.

> Do we need this line? The height should be resolved via getHeightForRow, and settings width to UIScreen.main.bounds.width shouldn't be required.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/BaseViewController.swift
```diff
@@ -0,0 +1,16 @@
...
+//  BaseViewController.swift
+//  SwiftSeedProject
+//
+//  Created by Lucas Pelizza on 4/21/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import UIKit
+
+class BaseViewController: UIViewController, BaseViewControllerProtocol {
```
mvazquez-ms, line 11: This doesn't seem to be used. (but I remember you used it before.. is this class still required?)

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/NewsFeedCollectionViewController.swift
```diff
@@ -0,0 +1,38 @@
...
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+
+import UIKit
+import Bond
+import Dip_UI
+import Dip
+
+private let reuseIdentifier = "NewsCollectionViewCell"
```
mvazquez-ms: This should be inside the class.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/NewsFeedCollectionViewController.swift
```diff
@@ -0,0 +1,38 @@
...
+
+class NewsFeedCollectionViewController: UICollectionViewController {
+
+    var viewModel: NewsFeedViewModel!
+    var dataSource: MutableObservableArray<MutableObservableArray<ArticleViewModel>>!
```
mvazquez-ms: Unless we'll have multiple sections, I'd recommend to use a single array instead of an array of arrays.

mvazquez-ms (commented):
Left some comments, most of them are fairly easy to fix.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/NewsFeedCollectionViewController.swift
briansztamfater: +1

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsCollectionViewCell.swift
```diff
@@ -0,0 +1,43 @@
...
+    static let Height = 80.0
+    
+    @IBOutlet weak var imgDetail: UIImageView!
+    @IBOutlet weak var lblDate: UILabel!
+    @IBOutlet weak var lblDescription: UILabel!
+    @IBOutlet weak var lblTitle: UILabel!
+    @IBOutlet weak var lblAuthor: UILabel!
+
+    
+    override func awakeFromNib() {
```
mvazquez-ms: You can (and should) remove this method completely now.

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/Components/NewsCollection/Cells/NewsCollectionViewCell.swift
```diff
@@ -0,0 +1,43 @@
...
+        super.awakeFromNib()
+    }
+    
+    func configure(viewModel: ArticleViewModel) {
+        lblDescription.text = viewModel.description
+        lblTitle.text = viewModel.title
+        lblAuthor.text = viewModel.author
+        lblDate.text = "\(viewModel.date)"
```
mvazquez-ms, line 28: What date format is used by default? Should we specify one?

mvazquez-ms (approved):
Good job! 👍

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/NewsFeedCollectionViewController.swift
```diff
@@ -0,0 +1,37 @@
...
+    var dataSource: MutableObservableArray<ArticleViewModel>!
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.collectionView!.register(NewsCollectionViewCell.self, forCellWithReuseIdentifier: reuseIdentifier)
+        bindViewModel()
+    }
+    
+    private func bindViewModel() {
+        dataSource = viewModel.newsArticles
```
briansztamfater: I think it would be better bind viewModel.newsArticles directly.

```
 viewModel.newsArticles.bind(to: collectionView!) { dataSource, indexPath, collectionView in
     let cell = collectionView.dequeueReusableCell(withReuseIdentifier: self.reuseIdentifier, for: indexPath) as! NewsCollectionViewCell
     cell.configure(viewModel: dataSource[indexPath.row])
     return cell
 }
```

What do you think?

IOS/SwiftSeedProject/SwiftSeedProject/Interfaces/ViewControllers/BaseViewController.swift
briansztamfater: +1

briansztamfater (approved):
Leaved some minor comments, but it is mergeable for me!

PR #20, opened 2017-04-21
[MLP-163] Refactor Rest Client

briansztamfater:
## Background

We need to refactor RestClient

## Changes done

- Refactored Endpoint to be now a Target protocol, so we can support multiple APIs
- Decoupled error sanitizing to handle it on each Target
- Same with custom headers, now each target can have its own custom headers
- Renamed some classes
- Made generic functions on RestClient so we don't repeat code
- Made some extensions for each endpoint so the code is better structured
- Created InfoUtils
- Reorganized classes in groups

## Approvers required

@mvazquez-ms @lpelizzaMS @mmaldini33


IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/Network/Protocols/RestClientManagerProtocol.swift
```diff
@@ -0,0 +1,23 @@
...
+protocol RestClientManagerProtocol {
+    func sendRequest(target: Target, parameters: [String : AnyObject]?, headers: [String : String]?) -> RestClientRequestProtocol
+}
+
+extension RestClientManagerProtocol {
+    func sendRequest(target: Target) -> RestClientRequestProtocol {
+        return sendRequest(target: target, parameters: nil, headers: nil)
+    }
+    
+    func sendRequest(target: Target, parameters: [String : AnyObject]?) -> RestClientRequestProtocol {
```
lpelizzaMS: maybe we can use this function always. Also, we can change this: 
func sendRequest(target: Target, parameters: [String : AnyObject]?) -> RestClientRequestProtocol
to this:
func sendRequest(target: Target, parameters: [String : AnyObject]? = nil ) -> RestClientRequestProtocol

lpelizzaMS (commented):
👍

IOS/SwiftSeedProject/SwiftSeedProject/Infrastructure/InfoUtils.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Created by Brian Sztamfater on 21/4/17.
+//  Copyright © 2017 Making Sense. All rights reserved.
+//
+
+import Foundation
+
+class InfoUtils {
+    
+    static func bundleNameKey() -> String {
+        guard let info = Bundle.main.infoDictionary,
```
mvazquez-ms, line 14: Watch out for the alignment here, it should look like this:

```Swift
guard let info = Bundle.main.infoDictionary,
      let domain = info[kCFBundleNameKey as String] as? String else {
    fatalError("Cannot get bundle name key from Info.plist")
}
```

(Golden Path section on our Style-guide)

mmaldini33 (commented):
Good changes!

mvazquez-ms (approved):
Looking good! Just left one minor comment about alignment.

PR #23, opened 2017-08-28
[MLP-191] Fix problem with Bond library

lpelizzaMS:
# Background
We should fix a problem with the Bond Library.

# Changes done
I've updated the Bond library version because I've found that the problem looked like there was an issue with CocoaPods not correctly setting SWIFT_ACTIVE_COMPILATION_CONDITIONS. So it is related to a problem reported on Bond git. This is the [Link](https://github.com/ReactiveKit/Bond/issues/411) to the issue.

# Notes
I've committed the Podfile.lock file because we need this for the CI.

# Approvers required
@mvazquez-ms @mmaldini33 @bsztamfater-ms

### DIFF
--- a/IOS/SwiftSeedProject/Podfile
+++ b/IOS/SwiftSeedProject/Podfile
@@ -9,7 +9,7 @@ target 'SwiftSeedProject' do
 
   pod 'Dip', '5.0.4'
   pod 'Dip-UI', '1.0.2'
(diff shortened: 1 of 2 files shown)


mmaldini33 (approved):
Ok

mvazquez-ms (approved):
👍

PR #26, opened 2018-02-19
[CAP-212] Creation of App using React-Native.

lpelizzaMS:
### Background

This is a first approach to the technology React Native. The app is called “Albums”. This app gets albums information from an endpoint Rest and shows the information on the app like a List.

###Features

- Show list of albums
- Redirect to Amazon to buy the album selected.

###Screenshot
rn-albums


ReactNative/Albums/src/components/AlbumList.js
```diff
@@ -0,0 +1,31 @@
...
+    console.log(this.state);
+
+    return (
+      <ScrollView>
+        {this.renderAlbums()}
+      </ScrollView>
+    );
+  }
+}
+
```
mravinale, line 30: good start!! 👍

mravinale: I think we can ignore all compiled files, and mantian only react-native code and configurations what do you think?

mmaldini33 (commented):
🥇 🥇  Nice job! Congratulations! 🥇 🥇

mravinale (approved):

mravinale:
let's go with redux! 
next PR :)

